PR #45, opened 2015-03-25
Allow to specify remote targets for deployments

maltewi:
``` Syskit.conf.use_deployments_from("aila_vision_deployments", :on => "vision_cpu")``` led to the following error:
```
= /home/sempromuser/RockDev/tools/syskit/lib/syskit/roby_app/plugin.rb:268:in `using_task_library': unknown options 'on' (ArgumentError)
= Backtrace
| 
|   /home/sempromuser/RockDev/tools/syskit/lib/syskit/roby_app/plugin.rb:268:in `using_task_library'
|   /home/sempromuser/RockDev/tools/syskit/lib/syskit/roby_app/configuration.rb:410:in `use_deployments_from'
|   ./models/profiles/aila.rb:29:in `<top (required)>'
|   /home/sempromuser/RockDev/tools/syskit/bin/syskit:19:in `<main>'
```

This this PR fixes the problem by adding ```:on => "localhost"``` to the optione evaluation in ```using_task_library``` and ```using_deployment```

### DIFF
--- a/lib/syskit/roby_app/configuration.rb
+++ b/lib/syskit/roby_app/configuration.rb
@@ -407,7 +407,7 @@ module Syskit
             # @see #use_deployment
             def use_deployments_from(project_name, options = Hash.new)
                 Syskit.info "using deployments from #{project_name}"
-                orogen = app.using_task_library(project_name, options)
+                orogen = app.using_task_library(project_name, options[:loader])
 
                 result = []
                 orogen.deployers.each_value do |deployment_def|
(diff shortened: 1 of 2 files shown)


doudou:
Since this on: option is unused in both methods, the right fix would be to not pass it there at all (i.e. modify use_deployments_from to pass only the :loader option to using_task_library)

doudou:
were things are ... being weird is that orogen_loaders was used on avalon, AFAIK, which also uses remote processes. /cc @saarnold, @goldhoorn

goldhoorn:
Jeah we used it during the SAUC-E, i dont know if we someshow had workarounded the problems...

https://github.com/auv-avalon/bundle/blob/master/config/avalon.rb

PR #46, opened 2015-03-26
Allow multiple devices per task

maltewi:
This is a reopening of #38 and #37. In https://github.com/rock-core/tools-syskit/pull/38#issuecomment-63317295 it was mentionen, that #37 would prevent component from running, because the `_dev` arguemnt would remain unset. This is now treated, by 6f2e1e3.
I'm using this patch for some time now and also `syskit run ...` is working fine.

### DIFF
--- a/lib/syskit/data_service.rb
+++ b/lib/syskit/data_service.rb
@@ -90,11 +90,7 @@ module Syskit
                     elsif driver_services.size > 1
                         raise ArgumentError, "#{self} handles more than one device, you must specify one of #{driver_services.map(&:name).sort.join(", ")} explicitely"
                     end
-                    binding.pry
-                    service = driver_services.find do |driver_service|
-                        service == driver_service
-                    end
-                    
+                    service = driver_services.first                    
                 end
 
                 arguments["#{service.name}_dev"]
--- a/lib/syskit/data_service.rb
+++ b/lib/syskit/data_service.rb
@@ -90,13 +90,14 @@ module Syskit
                     elsif driver_services.size > 1
                         raise ArgumentError, "#{self} handles more than one device, you must specify one of #{driver_services.map(&:name).sort.join(", ")} explicitely"
                     end
+                    binding.pry
                     service = driver_services.find do |driver_service|
                         service == driver_service
                     end
                     
                 end
 
-                arguments["dev"]
+                arguments["#{service.name}_dev"]
             end
 
             # Alias for #find_device_attached_to for user code
--- a/lib/syskit/data_service.rb
+++ b/lib/syskit/data_service.rb
@@ -90,10 +90,13 @@ module Syskit
                     elsif driver_services.size > 1
                         raise ArgumentError, "#{self} handles more than one device, you must specify one of #{driver_services.map(&:name).sort.join(", ")} explicitely"
                     end
-                    service = driver_services.first
+                    service = driver_services.find do |driver_service|
+                        service == driver_service
+                    end
+                    
                 end
 
-                arguments["#{service.name}_dev"]
+                arguments["dev"]
             end
 
             # Alias for #find_device_attached_to for user code
--- a/lib/syskit/models/component.rb
+++ b/lib/syskit/models/component.rb
@@ -616,7 +616,7 @@ module Syskit
             # Component.data_service for the description of +arguments+
             def driver_for(model, arguments = Hash.new, &block)
                 dserv = provides(model, arguments)
-                argument "#{dserv.name}_dev"
+                argument "dev"
                 dserv
             end
 
--- a/lib/syskit/models/component.rb
+++ b/lib/syskit/models/component.rb
@@ -616,7 +616,7 @@ module Syskit
             # Component.data_service for the description of +arguments+
             def driver_for(model, arguments = Hash.new, &block)
                 dserv = provides(model, arguments)
-                argument "dev"
+                argument "#{dserv.name}_dev", :default => nil
                 dserv
             end
 
--- a/lib/syskit/network_generation/dataflow_dynamics.rb
+++ b/lib/syskit/network_generation/dataflow_dynamics.rb
@@ -246,6 +246,10 @@ module Syskit
             def initial_device_information_common(task, triggering_devices)
                 triggering_devices.each do |service, device|
                     DataFlowDynamics.debug { "  #{device.name}: #{device.period} #{device.burst}" }
+                    if not device
+                        DataFlowDynamics.warn { "  No device, skipping" }
+                        next
+                    end
                     device_dynamics = PortDynamics.new(device.name, 1)
                     if device.period
                         device_dynamics.add_trigger(device.name, device.period, 1)
--- a/lib/syskit/network_generation/dataflow_dynamics.rb
+++ b/lib/syskit/network_generation/dataflow_dynamics.rb
@@ -223,7 +223,8 @@ module Syskit
                     if(task.find_device_attached_to(srv))
                         [srv, task.find_device_attached_to(srv)]
                     end
-                end
+                end.compact
+
                 DataFlowDynamics.debug do
                     DataFlowDynamics.debug "initial port dynamics on #{task} (device)"
                     DataFlowDynamics.debug "  attached devices: #{triggering_devices.map { |srv, dev| "#{dev.name} on #{srv.name}" }.join(", ")}"
--- a/lib/syskit/network_generation/dataflow_dynamics.rb
+++ b/lib/syskit/network_generation/dataflow_dynamics.rb
@@ -220,7 +220,9 @@ module Syskit
             # Adds triggering information from the attached devices to +task+'s ports
             def initial_device_information(task)
                 triggering_devices = task.model.each_master_driver_service.map do |srv|
-                    [srv, task.find_device_attached_to(srv)]
+                    if(task.find_device_attached_to(srv))
+                        [srv, task.find_device_attached_to(srv)]
+                    end
                 end
                 DataFlowDynamics.debug do
                     DataFlowDynamics.debug "initial port dynamics on #{task} (device)"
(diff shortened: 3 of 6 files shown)


lib/syskit/data_service.rb
```diff
@@ -90,10 +90,13 @@ def find_device_attached_to(service = nil)
...
                         raise ArgumentError, "#{self} handles more than one device, you must specify one of #{driver_services.map(&:name).sort.join(", ")} explicitely"
                     end
-                    service = driver_services.first
+                    service = driver_services.find do |driver_service|
+                        service == driver_service
+                    end
+                    
                 end
 
-                arguments["#{service.name}_dev"]
```
doudou: I miss an explanation about that change. Argument names are task-global, so all devices will use the same argument ???!?!?!?

lib/syskit/data_service.rb
```diff
@@ -90,10 +90,13 @@ def find_device_attached_to(service = nil)
                     elsif driver_services.size > 1
                         raise ArgumentError, "#{self} handles more than one device, you must specify one of #{driver_services.map(&:name).sort.join(", ")} explicitely"
                     end
-                    service = driver_services.first
+                    service = driver_services.find do |driver_service|
+                        service == driver_service
```
doudou: 'service' is always nil in this loop. I.e., it never matches. Moreover, since we are in a branch where driver_services.size == 1 AND service == nil, the matched service is indeed guaranteed to be driver_services.first

lib/syskit/network_generation/engine.rb
```diff
@@ -311,16 +311,6 @@ def find_selected_device_in_hierarchy(argument_name, leaf_task, requirements)
                 _, model, _ = leaf_task.requirements.resolved_dependency_injection.selection_for(nil, requirements)
                 if model && dev = model.arguments[argument_name]
                     return dev
-                else
-                    devices = Set.new
```
doudou: What about that one ????

lib/syskit/data_service.rb
maltewi: Yes I made that change to not have each a single argument for each device. Where there is on argument per device type, that leads to having unassigned arguments on the task. This in turn prevents the task from being started.
This change implies that in a concrete component network each concrete TaskContext can only be driver for one particular device. I think this is a good implication and also mirrors the behavior from before this PR (where a task context could only be driver for one device type).

lib/syskit/network_generation/engine.rb
maltewi: This prevents automatic selection of devices. This indeed has an influence on syskit's behavior as compared to before the PR.
I prefer it this way, but I guess opinions might differ.

lib/syskit/data_service.rb
doudou: > This change implies that in a concrete component network each concrete TaskContext can only be driver for one particular device. I think this is a good implication and also mirrors the behavior from before this PR (where a task context could only be driver for one device type).

It's not a good implication since we had (and probably will have again) drivers that could drive multiple devices at the same time. I originally made the feature of having multiple drivers declared per task for this particular purpose.

I don't understand the comment about the behaviour before the PR. The behaviour should be that a task context can be driver for an arbitrary number of devices of any types. The argument name is unique on a per-bound service (i.e. it is using the as: argument of driver_for).

lib/syskit/network_generation/engine.rb
doudou: > This prevents automatic selection of devices. This indeed has an influence on syskit's behavior as compared to before the PR.

And because of this, I would much prefer having it in a different PR.

lib/syskit/data_service.rb
maltewi: > The behaviour should be that a task context can be driver for an arbitrary number of devices of any types.
The behavior after the PR is the following: A TaskContext can be declared as being a driver for a numer of different devices:
```
JointControl::FakeJointDriverTask.driver_for BesmanTest::FakeDevices::JointsPositionDriver, :as => 'dummy_position'
JointControl::FakeJointDriverTask.driver_for BesmanTest::FakeDevices::JointsVelocityDriver, :as => 'dummy_velocity'
```

But then, a concrete TaskContext, once instanciated can only be either ```BesmanTest::FakeDevices::JointsVelocityDriver``` or ```BesmanTest::FakeDevices::JointsPositionDriver```. 

There is no problem, where there are two concrete TaskContext ```JointControl::FakeJointDriverTask```, representing different devices:
```
            device(BesmanTest::FakeDevices::JointsPositionDriver, :as => 'right_hand') \
                .joint_names(*$right_hand_joints) \
                .with_conf('right_hand', 'position').prefer_deployed_tasks(/right_hand/)
            device(BesmanTest::FakeDevices::JointsVelocityDriver, :as => 'torso') \
                .joint_names(*$torso_joints)\
                .with_conf('torso', 'velocity').prefer_deployed_tasks(/torso/)
```
for example is valid.

Does this example help?

doudou: > Yes I made that change to not have each a single argument for each device. Where there is on argument per device type, that leads to having unassigned arguments on the task. This in turn prevents the task from being started.

This can be fixed by setting 'nil' as default argument with

~~~
default: nil
~~~

Since syskit verifies that at least one argument is set, that would be fine

doudou: > Does this example help?

I understand the implication, yes. But I don't see it as a welcome change.

maltewi: I'm missing currently what functionality is lost by my changes. Can you give an example?

lib/syskit/network_generation/engine.rb
doudou: Just to clarify my position: I am not against disabling auto-selection of anything BUT it should IMO be optional. I start to use syskit to replace small ruby scripts, and auto-selection there is a blessing

maltewi: I see. Assuming we wanted to allow multiple devices per task. Let's think about options we have for auto-selection.

1) Automatically select the first device type that would match
2) Raise if there are multiple options

Other ideas? Do you have a preferred behavior?

lib/syskit/data_service.rb
doudou: To clarify things: the change of allowing starting a driver if at least one device is assigned is fine by me.

I have a problem with the renaming of the argument to 'dev'. You could before have one TaskContext be *at the same time* driver for two (or more) devices. It has been used (I added it explicitely because the feature was needed), so I would feel bad about removing it.

maltewi: Again, a small code example:
```
JointControl::FakeJointDriverTask.driver_for BesmanTest::FakeDevices::JointsDriver, :as => 'joints1'
JointControl::FakeJointDriverTask.driver_for BesmanTest::FakeDevices::JointsDriver, :as => 'joints2'
```
```
device(BesmanTest::FakeDevices::JointsDriver, :as => 'left_arm') \
                .with_conf('same_conf').prefer_deployed_tasks("driver_deployment")
            device(BesmanTest::FakeDevices::JointsDriver, :as => 'right_arm') \
                .with_conf('sample_conf').prefer_deployed_tasks("driver_deployment")
```
This example should be valid and result in selecting "driver_deployment" for both devices? I was not aware of this possiblity.

If the feature is needed and used, of cause we need to support it. I'll have a look at your proposed method with the default argument.

lib/syskit/network_generation/engine.rb
doudou: I'm always for raising if there is an ambiguity, i.e. more or less what you propose:
 - if there are no choices, then don't do anything
 - if there is more than one choice, raise
 - if there is exactly one choice, assign

lib/syskit/data_service.rb
doudou: > This example should be valid and result in selecting "driver_deployment" for both devices? I was not aware of this possiblity.

Provided that I remember how to tell Syskit how to disambiguate (we would have to assign devices to driver services explicitely)

doudou:
As for your other PRs ... I apologize for the way too long delay.

Looks good overall. Please squash all the commits into one ... and fix some style issues (see comments).

lib/syskit/data_service.rb
```diff
@@ -90,7 +90,7 @@ def find_device_attached_to(service = nil)
                     elsif driver_services.size > 1
                         raise ArgumentError, "#{self} handles more than one device, you must specify one of #{driver_services.map(&:name).sort.join(", ")} explicitely"
                     end
-                    service = driver_services.first
+                    service = driver_services.first
```
doudou: Remove the trailing space

lib/syskit/models/component.rb
```diff
@@ -616,7 +616,7 @@ def provides(model, arguments = Hash.new)
             # Component.data_service for the description of +arguments+
             def driver_for(model, arguments = Hash.new, &block)
                 dserv = provides(model, arguments)
-                argument "#{dserv.name}_dev"
+                argument "#{dserv.name}_dev", :default => nil
```
doudou: Use keyword argument style

~~~
argument name, default: nil
~~~

lib/syskit/network_generation/dataflow_dynamics.rb
```diff
@@ -220,8 +220,11 @@ def add_port_trigger(task, port_name, name, period, burst)
             # Adds triggering information from the attached devices to +task+'s ports
             def initial_device_information(task)
                 triggering_devices = task.model.each_master_driver_service.map do |srv|
-                    [srv, task.find_device_attached_to(srv)]
-                end
+                    if(task.find_device_attached_to(srv))
```
doudou: No parenthesis around the argument of if

lib/syskit/network_generation/dataflow_dynamics.rb
```diff
@@ -246,6 +249,10 @@ def initial_device_information(task)
             def initial_device_information_common(task, triggering_devices)
                 triggering_devices.each do |service, device|
                     DataFlowDynamics.debug { "  #{device.name}: #{device.period} #{device.burst}" }
+                    if not device
```
doudou: use ! instead of not (for style consistency)

doudou:
For your information, github does not notify when a PR branch gets updated ... You have to explicitely ping in a PR thread to get the other people to notice your updates.

PR #47, opened 2015-03-26
Allow compositions to be device drivers

maltewi:
Previously only single TaskContexts could be used as device drivers. This PR allows also the assignment of compositions as driver for a device.

Two changes have been made:
1) search for Components (not TaskContexts), that fullfill the device model in RobotDefinition.device
2) Allow to specify a use-statement for devices. The given dependency hash is applied in MasterDeviceIsntance.to_instance_requirements only for Compositions (not for task contexts)

### DIFF
--- a/lib/syskit/robot/master_device_instance.rb
+++ b/lib/syskit/robot/master_device_instance.rb
@@ -34,6 +34,8 @@ module Syskit
             attr_reader :driver_model
             # The task arguments
             attr_reader :task_arguments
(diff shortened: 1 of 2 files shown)


doudou:
Hi. I'm very sorry that I let this rot for so long ... I took on myself to improve it, did not feel like asking you after having let it bit-rot for what, 6 months ?

Result is at #69. Could you notify me to tell me if you plan to test it or not ?

PR #85, opened 2015-11-11
start actions with arguments using the syskit ide

gustavoneves12:


doudou:
Hi. I like the idea with the YAML, but unfortunately it works really well only with numbers and strings. If one expects a symbol or nil, it's not supported in a straightforward manner in YAML.

The idea I had was to basically use the text box as a Roby shell, i.e. one would enter the command as if entered in the shell and then we would parse the command in a batch context, e.g.

~~~
batch = client.create_batch
begin
   batch.instance_eval(command_string)
rescue Exception => e
   # Present error to user
end
client.process_batch(batch)
~~~

The pre-filled text box helps the user when the aim is starting a specific job, but one could also kill jobs or start multiple actions in parallel in this way.

One other thing: I would prefer that you keep separate the user input (the dialog box + parsing) from what is being done with it (the job start/restart). I really don't like the fact that the dialog connects to the syskit instance independently.

gustavoneves12:
@doudou 

Why do we need a parser? Would user inform only the syskit shell command in the dialog box?

E.g.
```
command_string = 'constant_z_def! :setpoint => -10'
```

If the user type something wrong, it throws a exception.

But, using the way which you suggested, how can I get restart the job?

doudou:
> Would user inform only the syskit shell command in the dialog box?

That's not what I meant (not running the full shell), but something LIKE the shell.

Moreover, there's no magic. When using instance_eval (which is basically what the shell does), you're "parsing" using the ruby interpreter itself. The nice thing about going through a batch is that the user can send multiple commands at once and that we're guaranteed that either no commands or all commands will be sent if there's an error in the text.

> But, using the way which you suggested, how can I get restart the job?

This is why I want to separate the interaction with the Client class with the user interaction.

Starting and restarting jobs are two different workflows. One can start the same job multiple times, which is why starting a new job should really care only about starting a new job [Syskit::GUI::RuntimeState#create_new_job](https://github.com/rock-core/tools-syskit/blob/master/lib/syskit/gui/runtime_state.rb#L325)

If you want to restart a job, you go on the job you want to restart and click "Start Again" or "Restart". This is handled [in the job UI itself](https://github.com/rock-core/tools-syskit/blob/master/lib/syskit/gui/job_status_display.rb#L92). There, instead of plainly calling JobMonitor#restart, you would need to
 - create a new job dialog, filling it in with the current job arguments
 - when the user presses "Go", add a kill_job command for the current job to the created batch and process the batch

I've written some [overview documentation about the interface API](https://github.com/rock-core/tools-roby/blob/d4720aa9e65736291f5f07d989c7abd897ea3f23/lib/roby/interface.rb). I hope it can help you.

PR #87, opened 2015-11-13
configure com busses to handle all their devices (even unattached ones) by default

doudou:
Reconfiguring a com bus is really something that should be avoided at
all costs. The current policy would cause a reconfiguration each time
a device on the bus would be activated for the first time, which would
disrupt the already running devices.

The default policy, after this commit, is to create all dynamic
services on the bus that are required for all declared devices. One can
get the earlier behaviour by adding lazy_dispatch: true to the com bus
declaration.

### DIFF
--- a/lib/syskit/data_service.rb
+++ b/lib/syskit/data_service.rb
@@ -184,23 +184,30 @@ module Syskit
             # Attaches the given task to the communication bus
             def attach(task)
                 model.each_com_bus_driver_service do |combus_srv|
+                    combus_m = combus_srv.model
+
                     # Do we have a device for this bus ?
                     next if !(combus = find_device_attached_to(combus_srv))
                     task.each_master_device do |dev|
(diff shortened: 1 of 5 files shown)


doudou:
@gustavoneves12: with that one, the can bus should not be reconfigured when you start the thrusters one by one. You should test it by starting and stopping thrusters, and checking the state of the CAN module.

gustavoneves12:
@doudou using the orogen/thrusters-enitech master the system breaks when we try to start one thruster. However, when we execute using configuration_hack_for_syskit branch the system keep working.

doudou:
did you check what happens if you start the thrusters one by one ? Is the CAN module reconfigured when you do so ?

doudou:
(You still need to use configuration_hack_for_syskit, the configure vs. start problem will be fixed by #88)

PR #95, opened 2015-12-10
Fix merge corner cases

doudou:
This fixes a few corner cases when deploying compositions. They were hit (but not caught) by the data monitor / fault table tests, in which it is common to monitor different outputs of the same network - one for monitoring and one for real usage.


doudou:
This is part of syskit2 ... no point.

PR #105, opened 2017-01-24
Run all blocking operations asynchronously

doudou:
This builds on top of [https://github.com/rock-core/tools-roby/pull/54](Roby async support) to run all network blocking operations asynchronously.

The only change in the public API is that Syskit got a real property API now, since properties are updated asynchronously as well. One needs to use the #properties accessor so:

```
properties.my_property = 10
```

instead of accessing the orocos_task directly. Note that this is a *must*, as the old way does not work anymore.

In addition, Syskit got a lot more / better testing assertions / harness.


doudou:
This has been tested extensively in simulation, and ran yesterday the whole day live.

PR #111, opened 2017-03-23
actions: pass more information when droby-marshalling/demarshalling an action

doudou:
On the user-visible side, this fixes displaying Syskit actions in the
roby shell.

Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/61


doudou:
@gustavoneves12

PR #112, opened 2017-03-24
Bugfixes

doudou:


doudou:
Tested in simulation and live today

PR #113, opened 2017-03-31
define MasterDeviceInstance#== and SlaveDeviceInstance#==

doudou:


doudou:
Fix confirmed by @gustavoneves12

PR #127, opened 2017-11-09
ide: add a gui to manage logging configuration

g-arjones:
So, I will do all the work here. Feel free to comment whenever you like.  I'm not sure what branch I should be basing this on (there are no shell interface tests at all on master :()

### DIFF
--- a/lib/syskit/gui/logging_configuration.rb
+++ b/lib/syskit/gui/logging_configuration.rb
@@ -1,0 +1,92 @@
+require 'vizkit'
+require 'vizkit/vizkit_items'
+require 'vizkit/tree_view'
+require 'Qt4'
+require 'syskit/shell_interface'
+require 'syskit/gui/logging_configuration_item'
+require 'roby/interface/exceptions'
+
+module Syskit
+    module GUI
+        # A widget containing an editable TreeView to allow the user to
+        # manage basic Syskit's logging configuration
+        class LoggingConfiguration < Qt::Widget
+            attr_reader :model, :treeView, :syskit, :item_name, :item_value
+            def initialize(parent = nil, syskit)
+                super(parent)
+                main_layout = Qt::VBoxLayout.new(self)
+                @treeView = Qt::TreeView.new
+
+                Vizkit.setup_tree_view treeView
+                @model = Vizkit::VizkitItemModel.new
+                treeView.setModel @model
+                main_layout.add_widget(treeView)
+                treeView.setColumnWidth(0, 200)
+                treeView.style_sheet = "QTreeView { background-color: rgb(255, 255, 219);
+                                                    alternate-background-color: rgb(255, 255, 174);
+                                                    color: rgb(0, 0, 0); }
+                                        QTreeView:disabled { color: rgb(159, 158, 158); }"
+
+                @syskit = syskit
+                @timer = Qt::Timer.new
+                @timer.connect(SIGNAL('timeout()')) { refresh }
+                @timer.start 10000
+
+                refresh
+            end
+
+            # Fetches the current logging configuration from syskit's
+            # sync interface
+            def refresh
+                if syskit.reachable?
+                    begin
+                        conf = syskit.client.call ['syskit'], :logging_conf
+                        update_model(conf)
+                        enabled true
+                    rescue Roby::Interface::ComError
+                        enabled false
+                    end
+                else
+                    enabled false
+                end
+            end
+
+            # Expands the entire tree
+            def recursive_expand(item)
+                treeView.expand(item.index)
+                (0...item.rowCount).each do |i|
+                    recursive_expand(item.child(i))
+                end
+            end
+
+            # Changes the top most item in the tree state
+            # and makes it update its childs accordingly 
+            def enabled(toggle)
+                @item_name.enabled toggle unless @item_name.nil?
+            end
+
+            # Updates the view model
+            def update_model(conf)
+                if @item_name.nil?
+                    @item_name = LoggingConfigurationItem.new(conf, :accept => true)
+                    @item_value = LoggingConfigurationItem.new(conf)
+                    @item_value.setEditable true
+                    @item_value.setText ""
+                    @model.appendRow([@item_name, @item_value])
+                    recursive_expand(@item_name)
+
+                    @item_name.on_accept_changes do |new_conf|
+                        begin
+                            conf = syskit.client.call ['syskit'], :update_logging_conf, new_conf
+                        rescue Roby::Interface::ComError
+                            enabled false
+                        end
+                    end
+                else
+                    return if @item_name.modified?
+                    @item_name.update_conf(conf)
+                end
+            end
+        end
+    end
+end
--- a/lib/syskit/gui/logging_configuration.rb
+++ b/lib/syskit/gui/logging_configuration.rb
@@ -88,7 +88,9 @@ module Syskit
 
                     @item_name.on_accept_changes do |new_conf|
                         begin
-                            conf = syskit.client.call ['syskit'], :update_logging_conf, new_conf
+                            syskit.async_call ['syskit'], :update_logging_conf, new_conf do |error, result|
+                                enabled false unless error.nil?
+                            end
                         rescue Roby::Interface::ComError
                             enabled false
                         end
--- a/lib/syskit/gui/logging_configuration.rb
+++ b/lib/syskit/gui/logging_configuration.rb
@@ -11,7 +11,7 @@ module Syskit
         # A widget containing an editable TreeView to allow the user to
         # manage basic Syskit's logging configuration
         class LoggingConfiguration < Qt::Widget
-            attr_reader :model, :treeView, :syskit, :item_name, :item_value
+            attr_reader :model, :treeView, :syskit, :item_name, :item_value, :pending_call
             def initialize(parent = nil, syskit)
                 super(parent)
                 main_layout = Qt::VBoxLayout.new(self)
@@ -30,9 +30,14 @@ module Syskit
                 @syskit = syskit
                 @timer = Qt::Timer.new
                 @timer.connect(SIGNAL('timeout()')) { refresh }
-                @timer.start 10000
+                @timer.start 1500
 
                 refresh
+            end
+
+            # Whether there is a refreshing call pending
+            def refreshing?
+                syskit.async_call_pending?(pending_call)
             end
 
             # Fetches the current logging configuration from syskit's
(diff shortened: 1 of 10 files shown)


g-arjones:
This is the shell interface I came up with.. It allows all the basic operations:
- enable/disable port/conf logging
- create groups
- remove groups
- enable/disable logging of groups

doudou:
Please base it on `syskit2`. Which has tests.

`master` is unfortunately grossly outdated.

g-arjones:
👍

lib/syskit/shell_interface.rb
```diff
@@ -170,6 +170,45 @@ def disable_log_group(string)
         def disable_logging_of(string)
             disable_log_group(string)
         end
+
+        LoggingGroup = Struct.new(:enabled)
```
doudou: I would suggest putting the name here as well. It usually helps later, or one ends up passing both the group object and the name at the same time to each functions.

lib/syskit/shell_interface.rb
```diff
@@ -170,6 +170,45 @@ def disable_log_group(string)
...
+                                     Syskit.conf.logs.disable_port_logging
+
+            conf.conf_logs_enabled ? Syskit.conf.logs.enable_conf_logging :
+                                     Syskit.conf.logs.disable_conf_logging
+
+            conf.groups.each_pair do |name, group|
+                begin
+                    Syskit.conf.logs.group_by_name(name).enabled = group.enabled
+                rescue ArgumentError
+                    Syskit.conf.logs.create_group(name, enabled: group.enabled)
```
doudou: This creates an empty group. Since there is no good way to create new groups for now, I would not allow creating or removing groups at all.

I would expect the following behavior:
- leave groups that are not mentioned as-is
- raise or ignore-and-warn for groups that do not exist

test/test_shell_interface.rb
```diff
@@ -151,6 +151,74 @@ module Syskit
...
+                end
+                Marshal.dump(conf)
+            end
+
+            it "makes a deep copy of the configuration" do
+                one = subject.logging_conf
+                two = one.deep_copy
+
+                assert (one == two)
+                assert !one.equal?(two)
```
doudou: This is `refute_same`

test/test_shell_interface.rb
```diff
@@ -151,6 +151,74 @@ module Syskit
...
+                    assert (group.enabled? == conf.groups[key].enabled)
+                end
+                Marshal.dump(conf)
+            end
+
+            it "makes a deep copy of the configuration" do
+                one = subject.logging_conf
+                two = one.deep_copy
+
+                assert (one == two)
```
doudou: This is `assert_equal`

lib/syskit/shell_interface.rb
```diff
@@ -170,6 +170,45 @@ def disable_log_group(string)
         def disable_logging_of(string)
             disable_log_group(string)
         end
+
+        LoggingGroup = Struct.new(:enabled)
+        LoggingConfiguration = Struct.new(:port_logs_enabled, :conf_logs_enabled, :groups) do
+            def deep_copy
```
doudou: You don't use #deep_copy anywhere.

test/test_shell_interface.rb
```diff
@@ -151,6 +151,74 @@ module Syskit
...
+                assert !one.equal?(two)
+            end
+
+            it 'triggers a redeployment when the configuration is updated' do
+                conf = subject.logging_conf
+                flexmock(subject).should_receive(:redeploy).once
+                subject.update_logging_conf(conf)
```
doudou: This tests that redeploy is called, not that the configuration got updated first.

lib/syskit/shell_interface.rb
g-arjones: I will in the gui..

lib/syskit/shell_interface.rb
g-arjones: Yeah, I thought about that too but then figured I could just leave the functionality there. I will remove it

g-arjones:
I think that is it..

doudou:
Looks good :+1:

doudou:
I don't know how you want to progress with this. We could merge this PR and then the UI, or merge both together. The good part of merging both together, is that we're then sure the API has everything you need (cue 'deep_copy').

g-arjones:
My plan was to merge everything together, unless the shell interface is of any use to you in which case I can open two PRs..

doudou:
Nope, everything together sounds good

g-arjones:
Since we are using the sync interface, I am refreshing the view @ a 10s interval since the logging configuration is not something very dynamic. It is a workaround I chose until we move to the async interface (the other option was to invoke a new thread and a new sync interface as sharing the same across different threads is probably not safe)

lib/syskit/gui/logging_configuration.rb
```diff
@@ -0,0 +1,92 @@
...
+                @timer.connect(SIGNAL('timeout()')) { refresh }
+                @timer.start 10000
+
+                refresh
+            end
+
+            # Fetches the current logging configuration from syskit's
+            # sync interface
+            def refresh
+                if !syskit.client.nil?
```
doudou: This is `if syskit.reachable?`. We unfortunately still have to use `syskit.client` directly in some cases, but I'm trying to remove that need as much as possible.

doudou:
Your remark about the sync/async calls got me wondering ... I've been looking harder at the call codepath. I'm puttting this here just to write down the stuff, not necessarily because I think it's necessary for this PR.

DRobyChannel#write_packet is non-blocking, so sending a call from the main thread is non-blocking. The only thing we would need to create an async `call` would be to wait for the replies asynchronously (which is in principle not so hard since one is meant to call `Async::Interface#poll` periodically).

*However*, the code currently assumes that there's only one call in flight at any moment. It would obviously be better to change this as well, but in the meantime we could stick to queueing pending calls until the return of the previous call has been received.

doudou:
I don't understand the reason behind the use of VizkitItem and RubyItem vs. a plain standard item model ... could you explain that to me ?

lib/syskit/gui/logging_configuration.rb
g-arjones: 👍  (I have squashed the patch for this one)

g-arjones:
> I don't understand the reason behind the use of VizkitItem and RubyItem vs. a plain standard item model ... could you explain that to me ?

I am using Vizkit's tree view extension and its acknowledge/enum editors and they use stuff from VizkitItem. I can copy those here if you prefer.. it is not a lot of code

g-arjones:
Regarding the remote calls, I was reading droby's codepath trying to think of something to fix that. My initial thoughts were:

1. Create new hooks to notify about changes in the logging configuration, or;
2. Create an api like `async_call do |response| ...; end` in `Roby::Interface::Async::Interface`

I could not find a good way to implement (1) in syskit because [I believe roby's client implementation blocks anything it doesn't recognize](https://github.com/rock-core/tools-roby/blob/master/lib/roby/interface/client.rb#L112) and doing that in roby itself doesn't make much sense IMO because it is very syskit specific. So, (2) sounds like the way to go for me and is probably more like what you are suggesting. This is already in my TODO list but it would be of a lot of help if you point me to where you believe this should go in roby's code base

g-arjones:
ping @doudou

lib/syskit/gui/logging_configuration.rb
```diff
@@ -40,9 +40,15 @@ def initialize(parent = nil, syskit)
             def refresh
                 if syskit.reachable?
                     begin
-                        conf = syskit.client.call ['syskit'], :logging_conf
-                        update_model(conf)
-                        enabled true
+                        return if syskit.client.pending_async_calls.size > 2
```
doudou: This can fail in many ways. When we start using `async_call` more often, we're bound to get more than 2 calls pending. In any case, it makes no sense to have more than one refresh call "in-flight" at any given time.

I would suggest

Either: to add an API on the Roby interface to test whether a particular call is queued (problem: it would not necessarily be "your call")

Or: to return an object from `async_call` that can be used to test whether the call is still "in-flight" (note that getting unreachable automatically resets  an object to not-in-flight").

Moreover, it would be good to get `#async_call` on the async interface, so that we don't access `syskit.client` directly.

g-arjones: > This can fail in many ways. When we start using async_call more often, we're bound to get more than 2 calls pending. In any case, it makes no sense to have more than one refresh call "in-flight" at any given time.

The point of this test was not to check for pending refresh calls per-se but any async call to avoid flooding roby. Since this is UI, if there are pending calls, just wait for those to be completed and only then refresh the UI. Kinda like assigning low priority to UI refreshing.

> return an object from async_call that can be used to test whether the call is still "in-flight"

This can be done in the block itself. 

```
return if refreshing?
refreshing = true
async_call(...) { |error, result| refreshing = false }
```

But again, this was not my concern when I added this check.

> Moreover, it would be good to get #async_call on the async interface, so that we don't access syskit.client directly.

Agree! I will propose something in roby..

doudou: > The point of this test was not to check for pending refresh calls per-se but any async call to avoid flooding roby.

Without a form of feedback - and IMO there's no good way to do feedback there, I'd rather "flood" roby than ignore the refresh. The limit of 2 is completely arbitrary (and as usual with these things, it would be hard to find a universal limit) and another async_call user could easily starve the refresh -- which would be totally confusing for a user.

IMO the best here is to ignore the possible "flooding" issue for now, and introduce more complex forms of queue management when necessary. I also usually find separating the "call object" from callback itself more elegant, since even simple state tracking like this one is usually also easily fucked up.

What I *do* think is important, is to avoid queueing the same message over and over again unnecessarily, because that's usually what ends up creating huge queues under load (compounding on the load ...)

Note: medium-term, the best would probably to start having a "debug" window that would gather important UI-specific statistics (queued async calls, queue length in write_package, time spent processing messages, time spent blocked in test processing, ...)

g-arjones: Fair enough.. Amended the last commit 👍

doudou:
> I can copy those here if you prefer.. it is not a lot of code

No need. I was just wondering. We already depend on Vizkit for the task inspector, so let's not nitpick. If we end up ripping out Vizkit from the IDE (very unlikely), we can copy the code at this point.

doudou:
Great work, thanks a lot.

FYI, there's a weird behavior of the task inspector I never noticed before. Reported here: https://github.com/rock-gazebo/gui-vizkit/issues/3

g-arjones:
No problem.. 👍

PR #140, opened 2017-11-22
fix dot's SVG output to not include < > in HREF's

doudou:
They are properly translated to &lt; and &gt; in the geneated dot, but
then dot converts them back to < and > in the SVG. This breaks SVG
parsing of e.g. inkscape.


doudou:
@alcantara09 will be directly useful for you given what you're doing ;-)

PR #150, opened 2018-01-26
implement an extension to Roby's REST API

doudou:
It for now provides deployment-management functionality suitable to implement a debugging interface for [the vscode extension](https://github.com/rock-core/vscode-rock)

Depends on:
- [ ] https://github.com/orocos-toolchain/orogen/pull/107
- [x] https://github.com/rock-core/tools-orocosrb/pull/118
- [x] https://github.com/rock-core/tools-roby/pull/82

Overrides for autoproj:

~~~
- orogen:
  branch: export_more_deployment_info_in_pkgconfig
- tools/orocos.rb:
  branch: extract_command_line_generation
- tools/roby:
  branch: rest_api
- tools/syskit:
  branch: rest_api
~~~


lib/syskit/roby_app/rest_api.rb
```diff
@@ -0,0 +1,181 @@
...
+                # Returns a string that describes the process server type
+                #
+                # @return [String,nil] the description, or nil if the given
+                #   process server should not be exposed by the REST API
+                def find_type_from_process_server(process_server_config)
+                    PROCESS_SERVER_TYPES[process_server_config.client.class]
+                end
+            end
+
+            get '/deployments/available' do
```
g-arjones: These APIs usually provide some kind of filtering.. IMO, filter deployments/tasks within a given project and deployments/tasks that contain task models defined in a given project would be very useful

lib/syskit/roby_app/rest_api.rb
```diff
@@ -0,0 +1,181 @@
...
+                        project_name: project_name,
+                        tasks: tasks.map { |task_name, task_model_name| Hash['task_name' => task_name, 'task_model_name' => task_model_name] },
+                        default_deployment_for: default_deployment_for,
+                        default_logger: default_logger
+                    ]
+                end
+                { deployments: info }
+            end
+
+            get '/deployments/registered' do
```
g-arjones: How about some documentation on these endpoints? :-)

lib/syskit/roby_app/rest_api.rb
doudou: There will never be that many available deployments. My policy there would be to filter at the client level.

Filtering in "real web apps" is usually present because (1) there can be a large number of responses and (2) it is easier on the database (gain when there's a lot of clients)

lib/syskit/roby_app/rest_api.rb
doudou: Docu ... what ? :P

doudou:
The problem with model loading led me to do some more refactoring there, and add some unit tests regarding the overall issue.

I still want to do the rock-gazebo tutorial to check that at least there's not something dramatically broken, but from the unit tests (and my understanding of the cleaned up setup sequence), it should be fine.

lib/syskit/roby_app/rest_api.rb
doudou: ... and done

g-arjones:
Looks all good now (tests are passing)

doudou:
And now I have to merge this ... or the app setup will break.

PR #151, opened 2018-01-29
ide: fix page subclass inheritance

g-arjones:
`syskit instanciate` sub command was broken due to wrong subclassing... @doudou double check this please


g-arjones:
@doudou ping

PR #154, opened 2018-02-08
fix corner cases in modelling and network generation

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/83

### DIFF
--- a/lib/syskit/graphviz.rb
+++ b/lib/syskit/graphviz.rb
@@ -1,4 +1,5 @@
 require 'roby/gui/dot_id'
+require 'syskit/gui/model_browser'
 
 module Syskit
         # Used by the to_dot* methods for color allocation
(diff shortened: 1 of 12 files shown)


g-arjones:
Looks alright.. What what the bug you mention in the commit log?

doudou:
> What was the bug you mention in the commit log?

Privately reported. The test basically repeats the bug:

- create a toplevel task with dynamic services and link it to a sink (in the bug's case, a logger)
- reconfigure that toplevel task, removing the connected port, and connect a new port to the logger
- run `fix_toplevel_task`. The replacement will merge the old connection with the new one and we end up with a network where a non-existing port is connected.

doudou:
I'm merging this now. It's been used "in production" for a while now.

PR #155, opened 2018-02-26
cli: fix main script

doudou:


g-arjones:
Looks hacky.. Isn't there a "thor way" of doing this?

doudou:
> Looks hacky.. Isn't there a "thor way" of doing this?

We would ideally move all the subcommands to thor, and then, at least temporarily, have thor start the scripts when subcommands are implemented through scripts. The only problem is `run`, which is a reserved command name and would need to be renamed into e.g. `start` or `exec`. However, I don't have the time to do this right now so I'm sticking with making this mixed thingy work.

g-arjones:
Fine :+1:

PR #161, opened 2018-03-08
droby: fix marshalling/demarshalling of ruby task contexts

doudou:


g-arjones:
Oh, well... Looks good. Just out of curiosity, how much of Roby's and Syskit's codebase is covered by unit tests? Also, is there are reason for travis not being enabled here?

doudou:
> Just out of curiosity, how much of Roby's and Syskit's codebase is covered by unit tests?

Line coverage is quite good (I don't have numbers, but the core code is above the 70-80% mark last time I looked, UI I don't know). Branch coverage is unfortunately a lot worse than this.

Now, this particular bug would probably not have been caught by unit test coverage, actually. The code *was* tested, just not this particular instance of it.

> Also, is there are reason for travis not being enabled here?

Unit-testing Syskit on travis would require to build a big chunk of the toolchain. Travis is *really* not meant for things that have dependencies. I have a Jenkins instance running tests at 13, but they don't run PRs (integrating the whole PR workflow within autoproj-jenkins would require quite a bit of time)

Roby has travis enabled ... but I was unfortunately less than good making sure it passes. Reference: https://github.com/rock-core/tools-roby/pull/93.

PR #170, opened 2018-04-11
ide: display the real type and not the intermediate type

doudou:
Apart from fixing confusing output, it also fixes the hyperlink of
a port to its type, after the removal of intermediate types from
the type list.


g-arjones:
:+1:

PR #175, opened 2018-04-24
runtime: allow tasks to specify that they can be started even if their inputs are not connected

doudou:
This is meant to reduce the ability for deadlocks, with tasks
like the logger that is synchronized with virtually everything.


doudou:
Ping @alcantara09

PR #186, opened 2018-06-15
Fix bug from disconnecting logger

alcantara09:
When mapping the deployed network to currently running tasks, it was a mistake in which tasks were added to the selected_deployments


doudou:
Hey ... fixed the issues we listed. I know you got tangled in EMMA's last tests ...

PR #194, opened 2018-11-06
Small test improvements

doudou:


lib/syskit/test/execution_expectations.rb
```diff
@@ -21,26 +21,48 @@ def self.resolve_orocos_reader(reader)
...
                 nil
             end
 
             # Expect that one sample arrives on the reader, and return the sample
             #
             # @return [Object]
             def have_one_new_sample(reader)
-                reader = ExecutionExpectations.resolve_orocos_reader(reader)
-                achieve(description: "receive one sample on #{reader}") { reader.read_new }
+                orocos_reader = ExecutionExpectations.resolve_orocos_reader(reader)
```
alcantara09: We should give the backtrace as argument to all the expectation, otherwise the failure backtrace is going to be the one of the nested expectation and not from the caller.

doudou: Done

PR #199, opened 2018-12-10
models: clean mixup between #resolve and #bind

doudou:
'bind' and 'resolve' are somewhat being used interchangeably in the codebase. This attemps to clean up the usage of both, with the following semantics.
    
'bind' "binds" a model and an already existing instance, returing the part of the instance that is representative of the model.
    
~~~
srv_m = DataService.new_submodel
task_m = TaskContext.new_submodel
task_m.provides srv_m, as: 'srv'
srv_m.bind(task_m) # => task_m.srv_m
~~~
    
'resolve' will more generally resolve relationships between objects. For instance, CompositionChild#resolve_and_bind_child will 'resolve' the child from a root composition and then bind it to the composition child model.

There is an inconsistency in the behavior of InstanceRequirements#bind, though, which always returns its argument (if it is valid), **not** providing a mapping between the model services and ports to the argument's actual ports.

However, since #bind is used for the return value of InstanceRequirements#instanciate, changing this would have far-reaching consequences that I can't deal with right now

I elected to keep the old behavior until I have the time to dig into it

### DIFF
--- a/test/models/test_data_services.rb
+++ b/test/models/test_data_services.rb
@@ -286,54 +286,67 @@ describe DataService do
     describe "#each_fullfilled_model" do
         it "includes the model itself, the service type and the root models" do
             parent_model = DataService.new_submodel
-            assert_equal [parent_model, DataService], 
+            assert_equal [parent_model, DataService],
                 parent_model.each_fullfilled_model.to_a
         end
         it "includes other service models it provides" do
             parent_model = DataService.new_submodel
             child_model  = DataService.new_submodel { provides parent_model }
-            assert_equal [child_model, parent_model, DataService], 
+            assert_equal [child_model, parent_model, DataService],
                 child_model.each_fullfilled_model.to_a
         end
     end
 
-    describe "#try_resolve" do
+    describe "#try_bind" do
+        before do
+            @srv_m = DataService.new_submodel
+            @task_m = Component.new_submodel
+        end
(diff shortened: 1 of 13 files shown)


test/models/test_data_services.rb
```diff
@@ -321,22 +321,30 @@ def test_it_can_be_droby_marshalled_and_unmarshalled
         end
     end
 
-    describe "#resolve" do
+    describe "#bind" do
         it "returns the value of try_resolve is non-nil" do
             srv_m = DataService.new_submodel
             flexmock(srv_m).should_receive(:try_resolve).with(task = flexmock).and_return(obj = flexmock)
```
alcantara09: should be better if everything was renamed to bind, including try_resolve => try_bind

doudou: This was a very astute observation, and a big rabbit hole.

I've update the commit (and the pull request) to clean up the usage of 'resolve' and 'bind' within the whole of Syskit. I spotted an issue with InstanceRequirements, and decided to leave it as it was as changing the behavior of IR#bind would IMO be far beyond the scope of this change (but will still need to be done)

doudou: I've updated the commit and PR messages to reflect the problem with InstanceRequirements

doudou:
There's this one too @alcantara09

PR #202, opened 2018-12-20
Deployment groups

doudou:
This PR is the last bit of code that was from FlatFish and had never reached mainline.

It is a major refactoring of the way deployments are managed, replacing the current global registration in Syskit.conf by an object oriented API at the profile level.

The goal of the PR itself is to get it merged, making sure that it doesn't break existing systems. I imagine that there would be further changes later to fine-tune the concept once we can start using it.


doudou:
@alcantara09 @elael @renan028 if you could simply apply the PR locally, to make sure that nothing gets broken by it, that would be great.

PR #203, opened 2018-12-24
Fix corner cases in child injection

doudou:


lib/syskit/models/composition_child.rb
```diff
@@ -39,6 +39,12 @@ def initialize_copy(old)
                 @overload_info = nil
             end
 
+            def freeze
```
alcantara09: @doudou I could not see where/how the freeze method is used. In the tests is there any case where it is called or the object is going to be frozen in a later usage?

doudou: InstanceRequirement objects are dup'ed and frozen when used in a `use()` call. This is to avoid having shared objects modified after the `use()`.

https://github.com/rock-core/tools-syskit/blob/master/lib/syskit/dependency_injection.rb#L121

and

https://github.com/rock-core/tools-syskit/blob/master/lib/syskit/dependency_injection.rb#L217

doudou: Ping @alcantara09

PR #207, opened 2019-10-24
refactor GenMain to factor out most of the code

doudou:
Depends on:
- [x] https://github.com/rock-core/tools-roby/pull/129

Also, rename templates to .erb to please vscode's syntax highlighting


g-arjones:
No tests?

doudou:
There are already a code-generation test for each type of element that can be generated (https://github.com/rock-core/tools-syskit/blob/master/test/cli/test_gen_main.rb). Thought adding unit-tests for the factored-out method won't add much (if anything).

PR #218, opened 2019-11-16
support nested namespaces in OroGenNamespace

doudou:
Once upon a time, a few years ago, Matthias implemented namespace
support in orogen. The main goal was to allow to define components
in a 'test' namespace, which would be disabled by default and only
enabled if one wants to run the orogen package's tests.

This gives access to tasks defined with OroGen's namespace
support through the OroGen toplevel namespace, e.g.
`OroGen.iodrivers_base.test.DriverTask` for `iodrivers_base::test::DriverTask`

### DIFF
--- a/lib/syskit/orogen_namespace.rb
+++ b/lib/syskit/orogen_namespace.rb
@@ -14,38 +14,52 @@ module Syskit
 
             def initialize(name)
                 @project_name = name
-                @registered_models = ::Hash.new
+                @registered_objects = {}
                 super()
             end
+
+            NAMESPACE_PREFIX = /([^:]+)::/.freeze
 
             def register_syskit_model(model)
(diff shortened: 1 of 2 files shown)


lib/syskit/orogen_namespace.rb
```diff
@@ -14,38 +14,52 @@ class ProjectNamespace < BasicObject
 
             def initialize(name)
                 @project_name = name
-                @registered_models = ::Hash.new
+                @registered_objects = {}
                 super()
             end
 
+            NAMESPACE_PREFIX = /([^:]+)::/.freeze
```
g-arjones: The regex allows both `:NAMESPACE` and `:::::::NAMESPACE`.. Is this desirable?

PR #222, opened 2019-12-05
cmake: configure the (orogen) test output to be compatible with autoproj if ROCK_TEST_LOG_DIR is set

doudou:


g-arjones:
Great 👍

PR #224, opened 2020-01-21
compositions: implement port_writer helper

g-arjones:
This is still missing error handling, documentation and tests. @doudou wanted to get your thoughts on the solution before investing time on that...

### DIFF
--- a/lib/syskit/models/composition.rb
+++ b/lib/syskit/models/composition.rb
@@ -33,10 +33,15 @@ module Syskit
                 Hash.new { |h, k| h[k] = [] }
             end
 
+            attribute(:port_writers) do
+                {}
+            end
+
             def clear_model
                 super
                 child_constraints.clear
                 children.clear
+                port_writers.clear
                 configurations.clear
                 exported_inputs.clear
                 exported_outputs.clear
@@ -350,6 +355,42 @@ module Syskit
                 end
 
                 @main_task = add(models, **options)
+            end
+
+            class PortWriter
+                attr_reader :child_name
+                attr_reader :port_name
+
+                def initialize(port)
+                    @child_name = port.component_model.child_name
+                    @port_name = port.name
+                end
+
+                def bind(cmp)
+                    @writer = cmp.child_from_role(child_name)
+                                 .find_input_port(port_name)
+                                 .writer
+                end
+
+                def write(sample)
+                    @writer.write(sample)
+                end
+            end
+
+            def port_writer_by_name(name)
+                @port_writers[name.to_sym]
+            end
+
+            def port_writer(name, port)
+                port_writers[name.to_sym] = PortWriter.new(port)
+
+                script do
+                    execute { model.port_writer_by_name(name.to_sym).bind(self) }
+                end
+
+                define_method name.to_sym do |sample|
+                    execute { model.port_writer_by_name(name.to_sym).write sample }
+                end
             end
 
             # Returns true if this composition model is a specialized version of


g-arjones:
Also, master is very outdated

lib/syskit/models/composition.rb
```diff
@@ -352,6 +357,42 @@ def add_main(models, **options)
...
+                                 .writer
+                end
+
+                def write(sample)
+                    @writer.write(sample)
+                end
+            end
+
+            def port_writer_by_name(name)
+                @port_writers[name.to_sym]
```
doudou: `@port_writers` will be nil here if called before `#port_writers` has ever been called. Just use the method version which will auto-initialize the 

Generally speaking, I'm not sure what this attribute brings ? The port could be resolved and memoized straight in the generated method.

g-arjones: > will be nil here if called before #port_writers has ever been called

Good point.. :+1: 

> I'm not sure what this attribute brings 

Seemed consistent with how everything else is stored. I can change to a plain instance variable though, no preference.

> The port could be resolved and memoized straight in the generated method.

In the generated method? And where would I keep the arguments passed in the model's `port_writer` helper (the port object, so far)

g-arjones:
> One thing that might work without going through the task script machinery, though, is to use Models::Composition#resolve_and_bind_child on the port's component model and then resolve the port, like so:
>
> component = model.resolve_and_bind_child_recursive(port.component_model)
> component.find_port(port.name)

The methods you mention don't exist in `Models::Composition`. They do in `Models::CompositionChild` though but since I don't have the child object in this context I am not sure how they can help.

Question: can I have two instances of the same service/component with same role (name) (in different compositions) or does syskit always enforce reuse of components?

doudou:
> The methods you mention don't exist in `Models::Composition`. They do in `Models::CompositionChild` though but since I don't have the child object in this context I am not sure how they can help.

Then add it there as well. It will allow resolving a child-or-root instead of only a child.

> Question: can I have two instances of the same service/component with same role (name) (in different compositions) or does syskit always enforce reuse of components?

Syskit will reuse if (1) the components have the same configuration and (2) the component's input streams are identical. I.e. it is possible that two instances of the same composition do not share the same component as child if one of these two conditions are not met.

g-arjones:
> Then add it there as well

I will have a look..

> it is possible that two instances of the same composition

Then I am not sure how what I proposed (using child and port name only to resolve the ports) can work reliably if we want to support recursion...What I mean is:

```ruby
class OneComposition < Syskit::Composition
    add(OroGen::Component::Task, as 'nested').with_conf('default', 'high')
end

class SecondComposition < Syskit::Composition
    add(OroGen::Component::Task, as 'nested').with_conf('other', 'low')
end

class OneComposition < Syskit::Composition
    add OneComposition, as 'a_foo'
    add SecondComposition, as 'another_foo'

    port_writer :foo_out, another_foo_child.nested_child.sensor_port
end
```

How would we prevent the engine from picking up `a_foo`'s root rather than the expected `another_foo`? Am I missing something?

g-arjones:
I guess that if OroGen::Component::Task was replaced with a composition added with different arguments in the different parents, the same would apply

doudou:
> Am I missing something?

`resolve_and_bind_child` starts from the root. You can have a child being part of two different compositions with the same role (i.e. parent selection from the child is ambiguous), but a given composition will always have a single child with a given name. Or I am the one missing something ...

g-arjones:
In the example above 'nested' isn't a child of root.. If we are looking for the port specification recursively in the children, we will then look for 'nested' in both 'a_foo' and 'another_foo' which both have matching ports..

doudou:
> In the example above 'nested' isn't a child of root.. If we are looking for the port specification recursively in the children, we will then look for 'nested' in both 'a_foo' and 'another_foo' which both have matching ports..

Yes. But `resolve_and_bind_child_recursive` starts from a root (in this case, the toplevel composition), and then resolves children recursively, even when called from a child-of-a-child.

See https://github.com/rock-core/tools-syskit/blob/master/lib/syskit/models/composition_child.rb#L85

doudou:
Unless I'm mistaken, this has been implemented by #241.

PR #230, opened 2020-03-16
app: allow configuring the name service used to resolve process server tasks

doudou:
This allows each remote system to be self-sufficient, instead of
relying on a root name service to be available.

### DIFF
--- a/test/roby_app/test_configuration.rb
+++ b/test/roby_app/test_configuration.rb
@@ -1,4 +1,5 @@
 require 'syskit/test/self'
+require 'syskit/roby_app/process_server'
 
 describe Syskit::RobyApp::Configuration do
     describe "#use_deployment" do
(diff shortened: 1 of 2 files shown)


test/roby_app/test_configuration.rb
```diff
@@ -65,4 +66,111 @@ def stub_deployment(name)
...
+            @mock_process = flexmock(pid: 20)
+            @mock_process.should_receive(:spawn).by_default
+            @mock_process.should_receive(:kill).by_default
+            @mock_process.should_receive(:dead!).by_default
+            @available_project_names = []
+            @available_typekit_names = []
+            @available_deployment_names = ['deployment']
+
+            set_log_level ::Robot, Logger::FATAL + 1
+            set_log_level Orocos::RemoteProcesses::Server, Logger::FATAL + 1
```
g-arjones: Are these hacks to disable logs?

doudou: It's to temporarily set the logger levels to avoid having them output stuff during the tests. They're reset on teardown.

PR #231, opened 2020-03-16
generically implement #deployed_as at the InstanceRequirements level

doudou:
It basically splits the logic between the generic and the specific
(the new #to_deployment_group call). This allows to implement it
as-is on RubyTaskContext as well.


g-arjones:
Looks good although the gymnastics around the #common_behavior test methods are weird.. Can't you just include the common tests as a module?

```ruby
module RubyTaskContextCommonBehavior
    extend Minitest::Spec::DSL

    it 'uses the default deployment using the given name' do
    end
end

describe RubyTaskContext do
    before do
    end

    after do
    end

    describe 'called on the task model' do
        before do
        end

        include RubyTaskContextCommonBehavior
    end

    describe 'called on InstanceRequirements' do
        before do
        end

        include RubyTaskContextCommonBehavior
    end
end
```

doudou:
This is the pattern I came up with a while ago.

One issue with the module is that it has to be defined at the toplevel. This is fine for this particular file, but would make the whole thing hard to follow for the much longer `test_task_context.rb`.

Basically, the alternative would be to create a custom helper that would return the module and then include that:

~~~
common_behavior = test_module do
   it ...
end

describe "called on the task model" do
   include common_behavior
end
~~~

If you think that it's much better, I can add that helper to Roby

g-arjones:
> If you think that it's much better, I can add that helper to Roby

I think that's better but I am not sure the helper would belong in Roby (maybe flexmock?). Also, not sure it's worth your time (probably not). If, in hindsight, you don't think it makes much difference, keep it as-is...

PS: It's kinda surprising Minitest itself doesn't offer a test inheritance mechanism... I think rspec does

doudou:
> PS: It's kinda surprising Minitest itself doesn't offer a test inheritance mechanism... I think rspec does

Yep. Quite a few blog posts spent on this subject over the years. Basically, the general sentiment is "it's rarely needed, and you can achieve the same using a plain Ruby construct"

The module helper thing would really belong into a minitest plugin. But I don't want to spent the time doing so, especially now that I've been using this method pattern a few times.

g-arjones:
> But I don't want to spent the time doing so, especially now that I've been using this method pattern a few times.

You are right... Given the circumstances, I think what you are using is fine

PR #241, opened 2020-04-09
Implement dynamic data sources

doudou:
Depends on:
- [x] https://github.com/rock-core/tools-roby/pull/153

Dynamic port bindings "anchor" a port dynamically in the system,
on the basis of e.g. a plan query, a port or a composition child. They
allow to resolve and re-resolve ports as actual components come
and go, and read from them

Their goal is to provide a service-oriented view of the data
sources in the plan, for instance for the purpose of a UI or monitoring.
Using dynamic data sources, one may tag specific compositions or
components with services that are then used to 'find' the appropriate
source at any given state, in a way that's orthogonal from the data flow.

The gist of the implementation is to use the refactored query API from Roby
(https://github.com/rock-core/tools-roby/pull/153) to provide a generalized port
query. From there, it's relatively simple machinery that allows to resolve data
service ports as well as plain component ports.

The user-facing API is at the component level, "simply":

~~~ ruby
data_reader ADataServiceModel.pose_samples_port, as: 'name'
# in the runtime code, just do name_reader.read / name_reader.read_new
data_reader some_child.port_name, as: 'name'
# in the runtime code, just do name_reader.read / name_reader.read_new
# Works for writers too
data_reader Services::TelemetryPublisher.stats_port, as: 'stats'
# in the runtime code, just do stats_writer.write
~~~


doudou:
@g-arjones, just realized that I've focussed on the read side, but the write side would be helpful as well. It's an easy extension, I'll take care of it before we merge. If you could still have a look, however, in case you have comments...

PR #246, opened 2020-05-08
 expand on the `have_.*_sample` riff

doudou:
Depends on:
- [x] https://github.com/rock-core/tools-roby/pull/164

- create the `.matching` refinement to the predicates, allowing to filter
  the samples against a predicate, for instance because one is looking
  for an (asynchronous) state change
- create `have_new_samples` to find more than one sample


g-arjones:
Cool... Thanks for this :+1:

PR #263, opened 2020-08-05
define deploy_current_plan

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/181

It runs the deployer on the complete plan. It uses run_planners,
and is therefore able to run method or state machine actions
before it checks for the deployment state.

It returns a mapping from the missions as they currently are to
the deployed missions


g-arjones:
Sorry for the delay, wanted to have a better look at this one

PR #269, opened 2020-12-02
fix the profile-level deployment group interface

doudou:
It was untested and full of bugs

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -115,7 +115,6 @@ Layout/LineLength:
     - 'syskit.gemspec'
     - 'test/actions/test_action_model.rb'
     - 'test/actions/test_interface_model_extension.rb'
-    - 'test/actions/test_profile.rb'
     - 'test/cli/test_gen_main.rb'
     - 'test/coordination/test_data_monitor.rb'
     - 'test/coordination/test_data_monitoring_table.rb'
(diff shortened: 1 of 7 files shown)


g-arjones:
I did not review this yet (I will) but just a heads up: I have two orogen tests failing in my CI because of both this and #268.
Didn't investigate yet so maybe it's not related to these changes but I found it kinda odd that it was the same two tests failing in both cases.

I suggest you run all your orogen tests with #268 and #269 and make sure they pass...

doudou:
> I suggest you run all your orogen tests with #268 and #269 and make sure they pass...

As far as I can see, I do have tests failing in CI ... but for Syskit itself. The orogen tests seem to pass. I just pushed the fixes for Syskit's unit tests, I'll double-check that the rest passes.

Maybe your tests were already failing ?

g-arjones:
> Maybe your tests were already failing ?

They were all green but they may be time-sensitive so it's probably just a coincidence since yours are passing...

doudou:
I can confirm that this one has no effect on our orogen component tests.

g-arjones:
I can confirm that mine breaks with this and doesn't without it. Still believe it's something with the test but I would like to fix them before you merge. Could you please keep this on hold for a couple of days until I figure it out?

doudou:
> Could you please keep this on hold for a couple of days until I figure it out?

Sure.

doudou:
For what it's worth, given the scope of the change here, it should break 100%, and at the plan generation level. Anything regarding runtime would IMO not make sense (i.e. could be a coincidence)

g-arjones:
All good. Thanks!

g-arjones (approved):

PR #282, opened 2021-03-30
spawn log transfer server in Application#setup

rbtmrcs:
Implements code to:
 
* Spawn a drop server that targets the app log directory within Application#setup
* Generate server certificate and password randomly at each start, and pass it to the created server
* Create a method to send a file transfer command to any given process server on RobyApp::Plugin, 
passing on the server URL, password and certificate as needed.

[work in progress]


lib/syskit/roby_app/plugin.rb
```diff
@@ -146,8 +148,64 @@ def self.setup(app)
...
+                if process_servers[name]
+                    # Establishes communication with said process server
+                    client = Syskit::RobyApp::RemoteProcesses::Client.new(
+                            host: process_server[name].host_id,
+                            port: @server_port)
+                    # Commands method log_upload_file from said process server
+                    client.log_upload_file(
+                        host: "localhost", 
+                        port: @log_transfer_port.port, 
+                        certificate: tmp_root_ca.cert_filepath,
```
doudou, line 174: Using a file here makes no sense. The process server is *remote*, it has no access to local files.

In other words, this should really be the certificate as a string, as you can see [here](https://github.com/rock-core/tools-syskit/blob/master/test/roby_app/remote_processes/test_remote_processes.rb#L192)

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,8 +148,64 @@ def self.setup(app)
...
+
+            def self.stop_tmp_root_ca
+                @tmp_root_ca = nil
+            end
+
+            def self.has_tmp_root_ca?
+                @tmp_root_ca
+            end
+
+            def self.start_local_log_transfer_server(tgt_dir, user, password, certificate)
```
doudou, line 196: Please separate paths from strings for the certificate arguments. Name paths with the `_path` suffix

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,8 +148,64 @@ def self.setup(app)
...
+                @log_transfer_server = Syskit::RobyApp::LogTransferServer::SpawnServer.new(tgt_dir, user, password, certificate)
```
doudou, line 197: Why assigning to instance variable here ? Just create. The instance variable will be stored by `.setup` (since its user will be `.cleanup`)

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,8 +148,64 @@ def self.setup(app)
...
+            end
+
+            def self.stop_local_log_transfer_server
+                @log_transfer_server = nil
```
doudou, line 201: Need to actually stop the server. Which would be given as argument by `.cleanup`

test/roby_app/test_plugin.rb
```diff
@@ -212,6 +212,22 @@ def perform_app_assertion(result)
...
+                after do
+                    Plugin.stop_tmp_root_ca
+                end
+                
+                it "has TmpRootCA been initialized" do
+                    assert_instance_of TmpRootCA, Plugin.has_tmp_root_ca?
+                end
+                
+            end
+
```
doudou, line 230: The test you can already write is (pseudocode):

~~~
root_ca = create_tmp_root_ca
start_log_transfer_server(root_ca)
# Do a transfer using send_file_transfer_command
# verify that the file transfer completes
~~~

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,8 +148,64 @@ def self.setup(app)
...
+                        user: tmp_root_ca.ca_user, 
+                        password: tmp_root_ca.ca_password, 
+                        localfile: logfile
+                    )
+                else
+                    raise ArgumentError, "there is no process server registered as #{name}"
+                end
             end
 
+            def self.start_tmp_root_ca
```
doudou, line 184: Since we have established that there is no stop for the CA, call this create_tmp_root_ca

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,8 +148,64 @@ def self.setup(app)
...
+                @tmp_root_ca = TmpRootCA.new
```
doudou, line 185: Would be better to save the instance variables in the setup call. Since "creating" the root CA is as simple as creating a TmpRootCA object, just remove the method altogether.

lib/syskit/roby_app/plugin.rb
```diff
@@ -876,6 +934,7 @@ def self.setup_rest_interface(app, rest_api)
                 require "syskit/roby_app/rest_api"
                 rest_api.mount REST_API => "/syskit"
             end
+
```
doudou, line 937: Stray line. Please execute rubocop before submitting code.

test/roby_app/spawn_server/test_spawn_server.rb
```diff
@@ -92,7 +92,7 @@ def upload_testfile
         it "uploads a file to the server's directory" do
             upload_testfile
             assert File.exist?("#{@temp_serverdir}/testfile"),
-                   "cannot find the expected upload"
+                "cannot find the expected upload"
```
doudou, line 95: This would fail in rubocop.

test/roby_app/spawn_server/test_spawn_server.rb
```diff
@@ -108,6 +108,7 @@ def upload_testfile
                     ftp.get("#{@temp_serverdir}/testfile")
                 end
             end
+            File.delete("testfile")
```
doudou, line 111: `@temp_serverdir` is created with `make_tmpdir`. It is therefore cleaned up automatically and you don't need to delete the file manually.

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,8 +148,64 @@ def self.setup(app)
...
+                )
+            end
+
+            def self.send_file_transfer_command(name, tmp_root_ca, logfile)
+                # Checks if said process server exists and is connected
+                if process_servers[name]
+                    # Establishes communication with said process server
+                    client = Syskit::RobyApp::RemoteProcesses::Client.new(
+                            host: process_server[name].host_id,
+                            port: @server_port)
```
doudou, line 169: To access process servers, use `Syskit.conf` (of type [Syskit::RobyApp::Configuration](https://www.rubydoc.info/github/rock-core/tools-syskit/master/Syskit/RobyApp/Configuration). In particular [process_server_config_for](https://www.rubydoc.info/github/rock-core/tools-syskit/master/Syskit/RobyApp/Configuration#process_server_config_for-instance_method) will check for existence and its `#client` attribute will return the client object, i.e.

~~~
client = Syskit.conf.process_server_config_for(name)
~~~

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,8 +148,64 @@ def self.setup(app)
...
+                    # Commands method log_upload_file from said process server
+                    client.log_upload_file(
+                        host: "localhost",
```
doudou, line 172: host cannot be `localhost`. `localhost` is always the machine __of the process that tries to create a connection__

In other words, this instructs the process server to connect to ftpd server that is on its own machine. Kind of defeats the purpose ...

Add the `host` parameter as a `#public_ip` property on Plugin (which is then accessible on the app object, that is Roby.app).  Do not start the log transfer server if it is not set.

lib/syskit/roby_app/plugin.rb
rbtmrcs: I've executed rubocop and corrected all offenses pointed, except this one.
It implies that the arguments are not needed, but without them it would raise an error of "0 of 4 arguments expected"
Screenshot_20210421_101629

rbtmrcs: Also, I've run rubocop in the TmpRootCA file and corrected most of offenses. There are a few remaining but they raise offenses as "Line is too long" and "Assingment Branch Condition size too high" (which I believe refers to the linecount of single methods). Do I correct them too or let them slide, been layout/style offenses?

lib/syskit/roby_app/plugin.rb
rbtmrcs: That reminded me a question I forgot to raise. 
How could we pass the certificate created by the TmpRootCA to the remotes process servers?
It came to mind creating an empty hash in the process server code to store that certificate, passed by the TmpRootCA
when created, but it was forgotten during the Niteroi's testing week.

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,6 +148,53 @@ def self.setup(app)
...
+
+                # Log Transfer FTP Server spawned during Application#setup
+                tmp_root_ca = TmpRootCA.new
+
+                start_local_log_transfer_server(
+                    app.log_dir,
+                    tmp_root_ca.ca_user,
+                    tmp_root_ca.ca_password,
+                    tmp_root_ca.signed_cert
+                )
```
doudou, line 160: Have to call stop_local_log_transfer_server in cleanup (need to define it)

~~~
def self.cleanup(app)
end
~~~

lib/syskit/roby_app/log_transfer_integration/log_transfer_integration.rb
```diff
@@ -0,0 +1,153 @@
...
+                    ef_cert.issuer_certificate = root_ca
+                    certificate_extension_add(ef_cert, cert)
+                    cert.sign(root_key, OpenSSL::Digest::SHA256.new)
+                    cert
+                end
+
+                # Write created Certificate to file
+                def write_certificate(cert)
+                    cert_filepath = File.join(__dir__,
+                        "..", "log_transfer_integration", "cert.crt")
```
doudou, line 133: If you want to write a certificate to a file, not passing the file path as parameter is rather obviously a Very Bad Idea(tm).

How does it work if I create two instances of TmpRootCA ?

Plus, writing in the tools/syskit package for an application is a VERY RED flag.

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,6 +154,50 @@ def self.setup(app)
...
+                client
+            end
+
+            def self.config_log_transfer
+                @log_transfer_port = @log_transfer_server.port
+                @log_transfer_certificate = @tmp_root_ca.certificate
+                @log_transfer_user = @tmp_root_ca.ca_user
+                @log_transfer_password = @tmp_root_ca.ca_password
+                ip = Socket.ip_address_list.detect(&:ipv4_private?)
+                @log_transfer_ip = ip.ip_address
```
doudou, line 186: - What I requested @rbtmrcs does is to make @log_transfer_ip a parameter that **must** be configured per-app. Do not try to guess that IP
- in any case, the purpose of log_transfer_ip is to be the IP from which a **remote** process server can access the FTP server. Picking a private IP is most probably not a very good default.

test/roby_app/test_plugin.rb
```diff
@@ -212,6 +212,143 @@ def perform_app_assertion(result)
...
+                        root_loader: root_loader
+                    )
+                end
+
+                def start_and_connect_to_server
+                    start_server
+                    connect_to_server
+                end
+
+                def start_process_server(name, host)
```
doudou, line 247: There are too many servers in these method names. I think this would be best named "register_process_server"

test/roby_app/test_plugin.rb
```diff
@@ -212,6 +212,143 @@ def perform_app_assertion(result)
...
+                    @ps_log_dir = make_tmpdir
+                    @client.create_log_dir(
+                        @ps_log_dir, Roby.app.time_tag,
+                        { "parent" => Roby.app.app_metadata }
+                    )
+                    Configuration.register_process_server(name, @client, log_dir, host_id: name)
+                end
+
+                def create_process_server(name)
```
doudou, line 256: Why the complexity ? The test_remote_process.rb test file was able to start a server and connect to it with a lot less setup code (which you copied but are not using here)

test/roby_app/test_plugin.rb
```diff
@@ -212,6 +212,143 @@ def perform_app_assertion(result)
...
+                end
+
+                it "uploads file from Process Server" do
+                    @ps_logfile = create_test_file(@ps_log_dir)
+                    path = File.join(app.log_dir, "logfile.log")
+                    refute File.exist?(path)
+                    @client = Plugin.send_file_transfer_command("test_ps", @ps_logfile)
+                    assert_upload_succeeds
+                    assert_equal File.read(path), File.read(@ps_logfile)
+                end
```
doudou, line 350: Would be better to put the actual test code first. The helpers are meant to be "implementation details"

doudou:
Last change:
- move all class-level methods and attributes related to the log transfer server (the `def self.` and `class << self`) to instance level
- make sure CI passes

For the last one, **before** you push, run

```
rake test
```

within `tools/syskit`. You need to have the test packages installed first, which can be done with

```
autoproj test enable tools/syskit
aup tools/syskit
```

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,6 +156,53 @@ def self.setup(app)
...
+                )
+                client
+            end
+
+            def self.config_log_transfer
+                @log_transfer_port = @log_transfer_server.port
+                @log_transfer_certificate = @tmp_root_ca.certificate
+                @log_transfer_user = @log_transfer_server.user
+                @log_transfer_password = @log_transfer_server.password
+            end
```
doudou, line 191: Instead of copying the value into separate attributes, just create separate accessor methods

```
def log_transfer_port
    @log_transfer_server.port
end
```

test/roby_app/test_plugin.rb
```diff
@@ -218,20 +218,20 @@ def perform_app_assertion(result)
...
-                    Plugin.log_transfer_ip = nil
+                    app.log_transfer_ip = nil
                 end
 
                 it "uploads file from Process Server" do
                     @ps_logfile = create_test_file(@ps_log_dir)
                     path = File.join(app.log_dir, "logfile.log")
                     refute File.exist?(path)
-                    @client = Plugin.send_file_transfer_command("test_ps", @ps_logfile)
+                    @client = app.send_file_transfer_command("test_ps", @ps_logfile)
```
doudou, line 234: What is this @client here ?

test/roby_app/test_plugin.rb
```diff
@@ -218,20 +218,20 @@ def perform_app_assertion(result)
...
                 end
 
                 after do
                     close_process_servers
-                    Plugin.log_transfer_ip = nil
+                    app.log_transfer_ip = nil
                 end
 
                 it "uploads file from Process Server" do
                     @ps_logfile = create_test_file(@ps_log_dir)
```
doudou, line 231: Do  you need to store that in an instance variable ? I don't see it used anywhere.

lib/syskit/roby_app/plugin.rb
```diff
@@ -158,51 +152,60 @@ def self.setup(app)
...
                 # Commands method log_upload_file from said process server
                 client.log_upload_file(
-                    @log_transfer_ip,
-                    @log_transfer_port,
-                    @log_transfer_certificate,
-                    @log_transfer_user,
+                    log_transfer_ip,
+                    log_transfer_port,
+                    log_transfer_certificate,
+                    log_transfer_user,
```
doudou, line 173: Seems to me that you don't actually need the attributes anymore. They could be private to the Plugin module (apart from log_transfer_ip). Is that right ?

test/roby_app/test_plugin.rb
MateusPedrosa: Do you mean having log_transfer_ip as an instance variable or specifically setting it to nil at that point?

MateusPedrosa: I think I get what you mean now. Do you think it should go back to Plugin?

lib/syskit/roby_app/plugin.rb
MateusPedrosa: That is right. I think we never really did, actually.

test/roby_app/test_plugin.rb
MateusPedrosa: send_file_transfer_command returns the client instance used to upload  the log so that we can check client.log_upload_state in wait_for_upload_completion

test/roby_app/test_plugin.rb
doudou: I meant @ps_logfile. Comments are applied to the last line of the code block.

test/roby_app/test_plugin.rb
doudou: Then, same comment than before. Do not pass arguments through instance variables. Give it as argument. I understand it's a side-effect of copy-pasting from the remote processes tests, but the situation there was different (the client was created far away in the setup blocks, in this case it is created and used right away)

test/roby_app/test_plugin.rb
MateusPedrosa: Oh okay

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,6 +150,58 @@ def self.setup(app)
...
+            def setup_local_log_transfer_server
+                @log_transfer_user = "process server"
+                @log_transfer_password = SecureRandom.base64(15)
+
+                tmp_root_ca = TmpRootCA.new(log_transfer_ip)
+                @log_transfer_certificate = tmp_root_ca.certificate
+
+                start_local_log_transfer_server(log_dir, @log_transfer_user, @log_transfer_password, tmp_root_ca.private_certificate_path)
+
+                @log_transfer_port = @log_transfer_server.port
```
doudou, line 167: Sorry ... Didn't explain myself properly.

You should *still* not define `@log_transfer_port` use `@log_transfer_server.port` whenever you need the port...

lib/syskit/roby_app/plugin.rb
```diff
@@ -146,6 +150,58 @@ def self.setup(app)
...
+                    tgt_dir,
+                    user,
+                    password,
+                    private_certificate_path
+                )
+            end
+
+            def stop_local_log_transfer_server
+                if @log_transfer_server
+                    @log_transfer_server&.stop
```
doudou, line 201: '&.' means "check if the receiver is nil, and return nil if it is instead of calling the method"

Since you check first that @log_transfer_server is not nil, '&.' is meaningless here.

doudou (approved):

PR #284, opened 2021-04-01
test: fix leaking of three I/Os per process server

doudou:
The process server is started in-process - mainly for speed -
and it turns out that the implementation was leaking a trap
and three I/Os. This is not an issue in production, as we
always start the servers in a separate process.

### DIFF
--- a/lib/syskit/roby_app/remote_processes/server.rb
+++ b/lib/syskit/roby_app/remote_processes/server.rb
@@ -135,7 +135,12 @@ module Syskit
 
                 def exec
                     open
-                    listen
+
+                    begin
+                        listen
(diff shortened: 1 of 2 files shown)


doudou:
For information, this is a shoot in the dark. I started to have random test failures in CI  related to the file transfer work (and therefore the integration of the process server implementation inside Syskit itself). Never managed to reproduce locally.

One difference I've seen before between docker in kubernetes in the cloud and local execution is the number of available file descriptors and available ports.... So I'm fixing the ones I can find.

doudou:
Found the actual source of the bug, workaround in #285

PR #285, opened 2021-04-02
Fix random bug related to ssl in ci

doudou:
Randomly - but rather often - CI would fail with "426 Connection closed; transfer aborted."

I could pinpoint the problem with the Net::FTP library closing the connection after the transfer is finished, but the remote side (ftpd) ending up getting a CONNRESET instead of a plain EOF. This would happen on SSL, I'm guessing because the SSL shutdown probably does a back-and-forth between client and server.

While I could reliably reproduce the bug on our CI's containers (Ubuntu 18.04, Ruby 2.5.1p57), I could not get it to trigger on my dev machine (Ubuntu 20.04, Ruby 2.7.0p0 and 2.7.1p83 compiled with rbenv). Even 2.5.0p0 and 2.5.1 do not trigger here.

So, there's something fishy here. I'm disabling TLS for the data connection for now, as it is unusable as it is. It's only half-bad as the ports are negotiated through the secure command connection, but we'll have to revisit the problem at some point

### DIFF
--- a/lib/syskit/roby_app/remote_processes/server.rb
+++ b/lib/syskit/roby_app/remote_processes/server.rb
@@ -135,7 +135,12 @@ module Syskit
 
                 def exec
                     open
-                    listen
+
+                    begin
+                        listen
+                    ensure
+                        close
+                    end
                 end
 
                 INTERNAL_SIGCHLD_TRIGGERED = "S"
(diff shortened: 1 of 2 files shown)


doudou:
@g-arjones I'm merging this now ... it's passing CI and the bug has been a hard block on our development process in a rather bad time. Of course, feel free to provide any comments you might have, I'll update the code accordingly.

doudou:
@rbtmrcs @MateusPedrosa this is of interest for both of you

PR #290, opened 2021-05-25
Fix reconfiguration logic

doudou:
On top of https://github.com/rock-core/tools-syskit/pull/289

The title sums it up ... it's embarrassing.

The reconfiguration logic was completely broken. Essentially, Syskit would always reconfigure, except if a component is configured with its default parameters. And also if it has no dynamic services.

The bad part is that fixing this required user-facing changes. This pull request require components to define a `update_properties` method in which the property updates are made. This method must be idempotent, and will be called in all cases. `configure` is called after the `cleanup` for non-idempotent configuration, as e.g. calling operations (the `canbus` component needs that for instance).

This change is forward and backward compatible though. Current orogen classes that do not define a configure method will be transparently using the new scheme. Tasks that do define `#configure` have to be migrated explictely to get the proper reconfiguration behavior, but will retain the current behavior if they do not.

This is documented by https://github.com/rock-core/rock-and-syskit/pull/72

### DIFF
--- a/lib/syskit/deployment.rb
+++ b/lib/syskit/deployment.rb
@@ -554,8 +554,14 @@ module Syskit
         # The currently applied configuration for the given task
         def configuration_changed?(orocos_name, conf, dynamic_services)
             current = remote_task_handles[orocos_name].current_configuration
-            current.conf != conf ||
-                current.dynamic_services != dynamic_services.to_set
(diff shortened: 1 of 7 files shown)


g-arjones (approved):

g-arjones:
The real surprise here is that no one noticed that before

PR #291, opened 2021-05-31
feat: implement rotation of data logs

rbtmrcs:
Depends on:
- [ ] https://github.com/rock-core/bundles-common_models/pull/36

Syskit implementation for triggering Log Rotation and Transference

Built on top of:
- [x] https://github.com/rock-core/tools-syskit/pull/282

### DIFF
--- a/lib/syskit/logger_service.rb
+++ b/lib/syskit/logger_service.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+module Syskit
+    task_service "LoggerService" do
+    end
+end
--- a/lib/syskit/logger_service.rb
+++ b/lib/syskit/logger_service.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# Decorator Service to enable log rotation
 module Syskit
     task_service "LoggerService" do
     end
--- a/lib/syskit/network_generation/logger.rb
+++ b/lib/syskit/network_generation/logger.rb
@@ -82,15 +82,26 @@ module Syskit
                 end
             end
 
-            @@logfile_indexes = {}
-
+            # From the orogen name "${logger_name}_Logger", removes the sufix "_Logger"
             def self.default_log_file_name(name)
                 name[/.*(?=_[L|l]ogger)/] || name
             end
 
-            def self.setup_default_logger(logger, log_file_name: default_log_file_name(logger.orocos_name))
-                index = (@@logfile_indexes[log_file_name] ||= -1) + 1
-                @@logfile_indexes[log_file_name] = index
+            def self.compute_next_log_index(previous_file)
+                unless previous_file
+                    return 0
+                end
+
+                next_index = previous_file[/.*(.[0-9]+.log)/, 1][/[0-9]+/].to_i + 1
+
+                next_index
+            end
+
+            def self.setup_default_logger(logger, logger_name: logger.orocos_name)
+                index = compute_next_log_index(logger.properties.file)
+
+                log_file_name = default_log_file_name(logger_name)
+
                 logger.properties.file = "#{log_file_name}.#{index}.log"
             end
 
--- a/lib/syskit/network_generation/logger.rb
+++ b/lib/syskit/network_generation/logger.rb
@@ -65,46 +65,10 @@ module Syskit
             def configure
                 super
 
-                if default_logger?
-                    deployment = execution_agent
-                    process = deployment.orocos_process
-                    process.setup_default_logger(
-                        self,
-                        log_file_name: process.default_log_file_name(orocos_task.basename),
-                        log_dir: deployment.log_dir,
-                        remote: !deployment.on_localhost?
-                    )
-                end
-
                 each_input_connection do |source_task, source_port_name, sink_port_name, policy|
                     source_port = source_task.find_output_port(source_port_name)
                     create_logging_port(sink_port_name, source_task, source_port)
                 end
-            end
-
-            def self.default_log_file_name(logger)
-                if logger.properties.file.nil? || logger.properties.file.empty?
-                    return logger.orocos_name.sub(/_[L|l]ogger/, "")
-                end
-
-                logger.properties.file[/.*(?=\.[0-9]+.log)/]
-            end
-
-            def self.compute_next_log_index(previous_file)
-                if previous_file.nil? || previous_file.empty?
-                    return 0
-                end
-
-                next_index = previous_file[/.*(.[0-9]+.log)/, 1][/[0-9]+/].to_i + 1
-
-                next_index
-            end
-
-            def self.setup_default_logger(logger)
-                index = compute_next_log_index(logger.properties.file)
-                file_name = default_log_file_name(logger)
-
-                logger.properties.file = "#{file_name}.#{index}.log"
             end
 
             def self.logger_dynamic_port
--- a/lib/syskit/network_generation/logger.rb
+++ b/lib/syskit/network_generation/logger.rb
@@ -83,7 +83,7 @@ module Syskit
             end
 
             def self.default_log_file_name(logger)
-                unless logger.properties.file
+                if logger.properties.file.nil? || logger.properties.file.empty?
                     return logger.orocos_name.sub(/_[L|l]ogger/, "")
                 end
 
@@ -91,7 +91,7 @@ module Syskit
             end
 
             def self.compute_next_log_index(previous_file)
-                unless previous_file
+                if previous_file.nil? || previous_file.empty?
                     return 0
                 end
 
--- a/lib/syskit/network_generation/logger.rb
+++ b/lib/syskit/network_generation/logger.rb
@@ -71,35 +71,20 @@ module Syskit
                 end
             end
 
-            def self.logger_dynamic_port
-                if @logger_dynamic_port
-                    return @logger_dynamic_port
-                end
-
-                ports = find_logger_model.orogen_model.dynamic_ports.find_all { |p| !p.type && p.kind_of?(OroGen::Spec::InputPort) }
+            def self.logger_dynamic_port_of(task_model)
+                ports = task_model.orogen_model.each_dynamic_input_port
+                                  .find_all { |p| !p.type }
                 if ports.size > 1
-                    raise InternalError, "oroGen's logger::Logger task should have only one catch-all dynamic input port"
+                    raise InternalError,
+                          "valid logger task contexts should have only one catch-all "\
+                          "dynamic input port, #{task_model} got #{ports.size}"
                 elsif ports.empty?
-                    raise InternalError, "oroGen's logger::Logger task should have one catch-all dynamic input port, and has none"
+                    raise InternalError,
+                          "valid logger task contexts should have one catch-all "\
+                          "dynamic input port, and #{task_model} has none"
                 end
 
-                @logger_dynamic_port = ports.first
-            end
-
-            def self.find_logger_model
-                TaskContext.find_model_from_orogen_name "logger::Logger"
-            end
-
-            def self.setup_logger_model(logger_model)
-                unless logger_model <= LoggerConfigurationSupport
-                    logger_model.include LoggerConfigurationSupport
-                    logger_model.stub do
-                        def createLoggingPort(port_name, port_type, metadata)
-                            create_input_port(port_name, port_type)
-                            true
-                        end
-                    end
-                end
+                ports.first
             end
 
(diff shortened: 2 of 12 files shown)


lib/syskit/logger_service.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Syskit
+    task_service "LoggerService" do
+        def initialize(**)
+            super
+        end
```
doudou, line 7: ? Why is that here ?

lib/syskit/roby_app/plugin.rb
```diff
@@ -250,6 +250,18 @@ def self.cleanup(app)
             # Hook called by the main application to prepare for execution
             def self.prepare(app)
                 @handler_ids = plug_engine_in_roby(app.execution_engine)
+
+                @log_upload_command_queue = Queue.new
+                @log_upload_thread = Thread.new { 
+                    app.log_upload_main(@log_upload_command_queue)
+                }
+                @rotated_logs = {}
```
doudou, line 258: Why is that a thread ? The process servers already handle this asynchronously, aren't they ?

lib/syskit/network_generation/logger.rb
```diff
@@ -82,6 +82,21 @@ def configure
                 end
             end
 
+            @@logfile_indexes = {}
+
+            def self.default_log_file_name(name)
+                name[/.*(?=_[L|l]ogger)/] || name
+            end
+
+            def self.setup_default_logger(logger, log_file_name: default_log_file_name(logger.basename))
```
doudou, line 91: `basename` is an artefact of orocos.rb. Within Syskit, just use `logger.orocos_name`

lib/syskit/network_generation/logger.rb
```diff
@@ -82,6 +82,21 @@ def configure
...
+                name[/.*(?=_[L|l]ogger)/] || name
+            end
+
+            def self.setup_default_logger(logger, log_file_name: default_log_file_name(logger.orocos_name))
+                index = (@@logfile_indexes[log_file_name] ||= -1) + 1
+                @@logfile_indexes[log_file_name] = index
+                log_file_path = "#{log_file_name}.#{index}.log"
+
+                logger.properties.file = log_file_path
+                logger
```
doudou, line 97: Just `logger.properties.file = "#{log_file_name}.#{index}.log"` would be sufficient

- the method does not have to return `logger` (if it should return something, it would be the log file name IMO)
- log_file_path is not used anywhere else, there's no real point creating a local variable.

lib/syskit/roby_app/plugin.rb
```diff
@@ -942,6 +957,32 @@ def self.setup_rest_interface(app, rest_api)
                 require "syskit/roby_app/rest_api"
                 rest_api.mount REST_API => "/syskit"
             end
+
+            def rotate_logs(rotated_logs)
+                plan.find_tasks(Syskit::LoggerService).running.each do |task|
+                    rotated_logs[task.execution_agent.on] = task.rotate_log
```
doudou, line 963: This assumes that there is only one task for any given remote process. Which is rather obviously wrong.

- write a separate test for this (which should fail)
- then fix the method

lib/syskit/roby_app/plugin.rb
```diff
@@ -942,6 +957,32 @@ def self.setup_rest_interface(app, rest_api)
...
+
+            def log_upload_main(log_upload_command_queue)
+                while (transfer = log_upload_command_queue.pop)
+                    transfer.apply
+                end
+            end
+
+            Upload = Struct.new(
+                :process_server_name, :file
+            ) do
```
doudou, line 981: Use this syntax to avoid crossing the 90 char line limit. For short lines, put everything on the same line. I don't have the impression that this one would be too long (I might be wrong though ...). Here, it would be

```
Upload = Struct.new(:process_server_name, :file) do
   ... 
end
```

test/roby_app/test_plugin.rb
```diff
@@ -313,6 +312,43 @@ def assert_upload_succeeds(client)
...
+                    @app.rotate_logs(@rotated_logs)
+                end
+
+                it "rotates log" do
+                    assert_equal @rotated_logs[@task.execution_agent.arguments[:on]], @logfile
+                end
+
+                it "verifies single process error in remote process server" do
+                    @rotated_logs.each do |key, value|
+                        assert(value.length() > 1, "identified single process per process server")
```
doudou, line 340: Don't simply check the count... Check the content !

lib/syskit/roby_app/plugin.rb
```diff
@@ -942,6 +957,51 @@ def self.setup_rest_interface(app, rest_api)
...
+
+            def rotate_logs(rotated_logs)
+                agents = get_running_agents(Syskit::LoggerService)
+                for agent in agents
+                    agent_logs = []
+                    plan.find_tasks(Syskit::LoggerService).running.each do |task|
+                        if task.execution_agent.arguments[:on] == agent
+                            task.rotate_log.each {|logfile| agent_logs << logfile}
+                        end
+                    end
```
caioaamaral, line 978: I don't get the logic here, is there a different plan for each agent iteration?

Also, looks like that `get_running_agents(Syskit::LoggerService)` is basically calling 
```rb
plan.find_tasks(Syskit::LoggerService).running.each
``` 
Why would we have to do it again inside the agents iteration, and why comparison `task.execution_agent.arguments[:on] == agent`? Isn't this always going to be true, or am I missing something?

test/network_generation/test_logger.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require "syskit/test/self"
-require "./test/fixtures/simple_composition_model"
+#require "./test/fixtures/simple_composition_model"
```
caioaamaral, line 4: why the comment?

lib/syskit/logger_service.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module Syskit
+    task_service "LoggerService" do
+    end
```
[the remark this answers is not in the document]
doudou: Because it's only used to "tag" the tasks that are logger services, so that the `find_tasks` work.

lib/syskit/roby_app/plugin.rb
```diff
@@ -264,6 +276,9 @@ def self.shutdown(app)
                     unplug_engine_from_roby(@handler_ids.values, app.execution_engine)
                     @handler_ids = nil
                 end
+
+                @log_upload_command_queue << nil
+                @log_upload_thread.join
```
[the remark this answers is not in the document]
doudou: I think it should, yes.

lib/syskit/roby_app/plugin.rb
```diff
@@ -250,6 +250,18 @@ def self.cleanup(app)
             # Hook called by the main application to prepare for execution
             def self.prepare(app)
```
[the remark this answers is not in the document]
doudou: `Roby::Application` from `tools/roby`

lib/syskit/roby_app/plugin.rb
doudou: You are absolutely right. It's a very complicated way, I think driven by the Mateus' need to group the logfiles per-process server (the :on argument is the process server name, `agent` is a misnomer. The execution agent is a task's deployment)

test/network_generation/test_logger.rb
doudou: No idea

lib/syskit/roby_app/plugin.rb
```diff
@@ -942,6 +957,51 @@ def self.setup_rest_interface(app, rest_api)
                 require "syskit/roby_app/rest_api"
                 rest_api.mount REST_API => "/syskit"
             end
+
+            def get_running_agents(syskit_task)
+                agents = []
+                plan.find_tasks(syskit_task).running.each do |task|
+                    agents << task.execution_agent.arguments[:on]
+                end
+                agents&=agents
```
caioaamaral, line 966: why the ampersand operator assignment on `agents` against itself?

lib/syskit/roby_app/plugin.rb
```diff
@@ -938,6 +953,51 @@ def self.setup_rest_interface(app, rest_api)
...
+            def rotate_logs(rotated_logs)
+                agents = get_running_agents(Syskit::LoggerService)
+                for agent in agents
+                    agent_logs = []
+                    plan.find_tasks(Syskit::LoggerService).running.each do |task|
+                        if task.execution_agent.arguments[:on] == agent
+                            task.rotate_log.each {|logfile| agent_logs << logfile}
+                        end
+                    end
+                    rotated_logs[agent] = agent_logs
```
caioaamaral, line 975: why we are bookkeeping the `rotated_logs`?

test/roby_app/test_plugin.rb
caioaamaral: does the "rotates log" test case already satisfy the content checking?

lib/syskit/roby_app/plugin.rb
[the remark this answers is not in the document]
caioaamaral: done

lib/syskit/roby_app/plugin.rb
```diff
@@ -938,6 +954,51 @@ def self.setup_rest_interface(app, rest_api)
...
+
+            def rotate_logs(rotated_logs)
+                tasks_book = get_tasks_per_process_server(Syskit::LoggerService)
+                tasks_book.each do |process_server, tasks|
+                    process_server_logs = []
+                    tasks.each do |task|
+                        task.rotate_log.each { |logfile| process_server_logs << logfile}
+                    end
+
+                    rotated_logs[process_server] = process_server_logs
```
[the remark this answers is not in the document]
caioaamaral: nice!

lib/syskit/network_generation/logger.rb
```diff
@@ -82,6 +82,31 @@ def configure
...
+                end
+
+                next_index = previous_file[/.*(.[0-9]+.log)/, 1][/[0-9]+/].to_i + 1
+
+                next_index
+            end
+
+            def self.setup_default_logger(logger)
+                index = compute_next_log_index(logger.properties.file)
+                file_name = default_log_file_name(logger)
```
doudou, line 105: Just a reminder, you did not change this line as we discussed.

caioaamaral: On the most cases I'm using the previous file (instead of the logger name), but I need to use something else when there is no previous file

doudou: > On the most cases I'm using the previous file (instead of the logger name)

I don't see that code ... where is it ?

If there are two codepaths (there are), you should make two methods. Once the component is configured, file is set. I.e. in `#update_properties` file is empty and in `#rotate_log` it is not.

There is actually a corner case that we don't handle right now. If a logger on which we did call rotate_log is stopped and started again, the index will be reset, since we don't update it in the [orocos.rb setup_default_logger](https://github.com/rock-core/tools-orocosrb/blob/master/lib/orocos/process.rb#L258).

Right now, I'm thinking that the best would be to move the all the code responsible to initialize and modify `file` in the default logger case (from LoggerSupport's `#configure`, orocos.rb as well as compute_next_index and friends) into bundles/common_models/orogen/logger.rb. What do you think ?

caioaamaral: > If there are two codepaths (there are), you should make two methods. Once the component is configured, file is set. I.e. in #update_properties file is empty and in #rotate_log it is not.

In some cases it is nil (test/network_generation/test_logger.rb:179), not sure why

> Right now, I'm thinking that the best would be to move the all the code responsible to initialize and modify file in the default logger case (from LoggerSupport's #configure, orocos.rb as well as compute_next_index and friends)

It seems great

From "orocos.rb" should we move only the `#setup_default_logger`, or should we include `#default_logger_name`, `#default_log_file_name`, etc?

caioaamaral: done in [ae5cd10](https://github.com/rock-core/tools-syskit/pull/291/commits/ae5cd109faa505cbeabb7c3abce3cda7258b9155)

doudou (approved):

caioaamaral:
ci looks good https://buildbot.tidewise.io/#/builders/1340/builds/186

let me just check if I can merge https://github.com/rock-core/bundles-common_models/pull/36

PR #292, opened 2021-06-14
test: fix have_no_new_sample/have_new_samples for boolean ports

doudou:
They return 'false' which was interpreted by the implementation as
"no new sample"


g-arjones (approved):

doudou:
@g-arjones if you could have another look. There were bugs lurking in the predicates implementation regarding buffer handling ... I fixed them in c2c1b40

g-arjones (approved):

PR #298, opened 2021-09-03
fix(app): re-enable export_types and syskit_model_constant_registration for backward compatibility

doudou:
While new apps do not need to be backward compatible, we auto-generate
apps for orogen tests, which have to be. Mention the two configuration
variables, but set them to true for now


doudou:
Merging this now ... Pretty urgent fix from #296

PR #315, opened 2022-03-17
fix: Enforce 'ruby_tasks autogeneration inside the './tasks' folder

caioaamaral:
Currently, `syskit gen ruby_task <task>`, generates the new task inside the `./models/compositions` and `./test/compositions` directory. 

With this change, the auto generated files will be in the `tasks` basename.


g-arjones:
Did @caioaamaral join TideWise? Cool!

PR #321, opened 2022-05-11
feat: add 'read_only' attribute for TaskContext

caioaamaral:
read-only tasks are tasks that share a resource (i.e "#orocos_task") with another task running on a different Syskit instance.
In that way, it does not interfere with the resource, as it will only perform read operations, skip setup, and also not stop when interrupted (so child tasks are not garbage collected).

### DIFF
--- a/lib/syskit/task_context.rb
+++ b/lib/syskit/task_context.rb
@@ -44,6 +44,8 @@ module Syskit
         # The name of the remote task context, i.e. the name under which it
         # can be resolved by Orocos.name_service
         argument :orocos_name
+        # Wether the task is set only for read operations
+        argument :read_only, default: false
 
         # [Orocos::TaskContext,Orocos::ROS::Node] the underlying remote task
         # context object. It is set only when the task context's deployment
@@ -70,6 +72,8 @@ module Syskit
             @pending_exception_states = []
 
             @calculated_dynamics = {}
+
+            return ready_to_start! if read_only?
 
             remote_handles.default_properties.each do |p, p_value|
                 syskit_p = property(p.name)
@@ -161,6 +165,13 @@ module Syskit
         # *and* all its inputs are connected
         def executable?
             @executable || (@ready_to_start && super)
+        end
+
+        # Whether this task context is read_only
+        #
+        # By default, its false
+        def read_only?
+            read_only
         end
 
         # Whether the task should be kept in plan
@@ -384,6 +395,15 @@ module Syskit
         # general property update structure, all property updates happening
         # in a single execution cycle will be committed together.
         def queue_property_update_if_needed
+            if read_only?
+                property = each_property.map { |p| p if p.needs_commit? }
+                                        .first
+
+                raise InvalidReadOnlyOperation,
+                      "attempting to write on property '#{property.name}',"\
+                      "but it's a 'read_only' Task"
+            end
+
             unless would_use_property_update?
                 raise InvalidState,
                       "attempting to queue a property update on a finished "\
@@ -693,7 +713,7 @@ module Syskit
         #  end
         #
         def setup?
-            @setup
+            read_only? || @setup
         end
 
         def ready_to_start!
@@ -913,7 +933,9 @@ module Syskit
                 properties_updated_in_configure =
                     properties.each.any?(&:needs_commit?)
             end
+
             commit_properties(promise)
+
             promise.then(description: "#{self}#perform_setup#orocos_task.configure") do
                 state = orocos_task.rtt_state
                 if properties_updated_in_configure && state != :PRE_OPERATIONAL
@@ -955,6 +977,9 @@ module Syskit
         # configured and started.
         event :start do |_context|
             info "starting #{self}"
+
+            return if read_only?
+
             @last_orogen_state = nil
 
             state_reader.resume if state_reader.respond_to?(:resume)
@@ -1151,7 +1176,9 @@ module Syskit
 
         # Interrupts the execution of this task context
         event :stop do |_context|
-            interrupt!
+            return interrupt! unless read_only?
+
+            stop_event.emit
         end
 
         def quarantined!(reason: nil)
--- a/lib/syskit/exceptions.rb
+++ b/lib/syskit/exceptions.rb
@@ -4,6 +4,7 @@ module Syskit
     class InternalError < RuntimeError; end
     class ConfigError < RuntimeError; end
     class SpecError < RuntimeError; end
+    class InvalidReadOnlyOperation < RuntimeError; end
 
     class Ambiguous < SpecError; end
 
--- a/test/models/test_task_context.rb
+++ b/test/models/test_task_context.rb
@@ -362,7 +362,7 @@ module Syskit # :nodoc:
 
             it "passes the :task_arguments option as arguments to the newly created task" do
                 task = task_model.instanciate(plan, DependencyInjectionContext.new, task_arguments: { conf: ["default"] })
-                assert_equal Hash[conf: ["default"]], task.arguments
+                assert_equal Hash[conf: ["default"], read_only: Roby::DefaultArgument.new(false)], task.arguments
             end
             it "sets the fullfilled model properly" do
                 arguments = Hash[conf: ["default"]]
--- a/test/test_task_context.rb
+++ b/test/test_task_context.rb
@@ -2240,5 +2240,88 @@ module Syskit
                 assert_equal "#{Process.pid}b", task.orocos_name
             end
         end
+
+        describe "read_only" do
+            before do
+                task_m = TaskContext.new_submodel do
+                    property "p", "/double"
+                end
+
+                @task =
+                    syskit_stub_deploy_and_configure(task_m.with_arguments(read_only: true))
+
+                Orocos.allow_blocking_calls { @task.orocos_task.configure(false) }
+            end
+
+            it "raises when attempting to change a property" do
+                task = @task
+
+                assert_raises(InvalidReadOnlyOperation) do
+                    task.properties.p = 2.0
+                end
+            end
+
+            it "emits start when the task is started while the component is running" do
+                task = @task
+                Orocos.allow_blocking_calls { task.orocos_task.start(false) }
+
+                assert state(task.orocos_task) == :RUNNING
+
+                expect_execution { task.start! }
+                    .to { emit task.start_event }
+            end
+
+            it "does not emit start" \
+            "if the task is started while the component is not running" do
+                task = @task
+
+                assert state(task.orocos_task) != :RUNNING
+
+                expect_execution { task.start! }
+                    .to { not_emit task.start_event }
+
+                # just to avoid: "TeardownFailedError: failed to tear down plan"
+                execute { task.stop! }
+            end
+
+            it "emits start when the component is started while the task is starting" do
+                task = @task
+                execute { task.start! }
+
+                assert task.starting?
+
+                expect_execution do
+                    Orocos.allow_blocking_calls { task.orocos_task.start(false) }
+                end.to { emit task.start_event }
+            end
+
+            it "does not emit interrupted" \
+            "if the task is stopped while the component is running" do
+                task = @task
+                Orocos.allow_blocking_calls { task.orocos_task.start(false) }
+                execute { task.start! }
+
+                assert task.running? && state(task.orocos_task) == :RUNNING
+
+                expect_execution { task.stop! }
+                    .to { not_emit task.interrupt_event }
+            end
+
+            it "emits stop when the component is stopped while the task is running" do
+                task = @task
+                Orocos.allow_blocking_calls { task.orocos_task.start(false) }
+                execute { task.start! }
+
+                assert task.running?
+
+                expect_execution do
+                    Orocos.allow_blocking_calls { task.orocos_task.stop(false) }
+                end.to { emit task.stop_event }
+            end
+
+            def state(component)
+                Orocos.allow_blocking_calls { component.rtt_state }
+            end
+        end
     end
 end


lib/syskit/task_context.rb
```diff
@@ -384,6 +397,11 @@ def would_use_property_update?
         # general property update structure, all property updates happening
         # in a single execution cycle will be committed together.
         def queue_property_update_if_needed
+
+            unless needs_setup?
+                raise "attempting to write a property on a 'read_only' Task"
```
doudou, line 402: Please define a proper exception class.

lib/syskit/task_context.rb
```diff
@@ -896,6 +914,12 @@ def prepare_for_setup(promise)
 
         # (see Component#perform_setup)
         def perform_setup(promise)
+            unless needs_setup?
```
doudou, line 917: The goal is to never be entering `perform_setup`

I was thinking that the simplest way would be to return true in `setup?` for readonly tasks (i.e. pretend they are set up)

lib/syskit/task_context.rb
```diff
@@ -959,6 +983,14 @@ def setup_failed!(exception)
 
             state_reader.resume if state_reader.respond_to?(:resume)
 
+            unless needs_setup?
```
doudou, line 986: **Here** is why `needs_setup?` is plain wrong. Why does `setup` (which relates to configuration) affects start ?

Just use read_only directly

lib/syskit/task_context.rb
```diff
@@ -959,6 +983,14 @@ def setup_failed!(exception)
...
+                promise = promise(description: "promise:#{self}#start")
+                commit_properties(promise)
+                promise.then { orocos_task.start(false) }
+                start_event.achieve_asynchronously(promise, emit_on_success: false)
```
doudou, line 990: Well. In read_only mode, we should ... not start the component.

Just return is fine.

test/test_readonly_task.rb
```diff
@@ -0,0 +1,31 @@
...
+
+            assert_raises(RuntimeError) do
+                task.properties.p = 2.0
+            end
+
+            syskit_configure(task)
+
+            expect_execution { task.start! }
+                .to { emit task.start_event }
+        end
```
doudou, line 29: I know it's a draft, but I think you could copy/paste the four tests I wrote on the chat, just to make sure you won't forget one ...

lib/syskit/task_context.rb
caioaamaral: Well, I was thinking about using `perform_setup` just for calling `setup_successful!` as otherwise calling `task.start!` would raise an exception.

The same applies for `orocos_task.configure(false)`, which I need in order to transition from `: PRE_OPERATIONAL` to `:STOP` (and then `task.start!` would emit the start_event).

lib/syskit/task_context.rb
```diff
@@ -955,6 +980,12 @@ def setup_failed!(exception)
         # configured and started.
         event :start do |_context|
             info "starting #{self}"
+
+            if read_only?
+                event(:start).emit
```
doudou, line 985: Nope. You need to just ... return

The start event should be emitted only if the C++ task context is running. update_task_states.rb should take care of that.

lib/syskit/task_context.rb
```diff
@@ -1089,6 +1120,11 @@ def stop_orocos_task
         event :interrupt do |_context|
             info "interrupting #{self}"
 
+            if read_only?
+                event(:interrupt).emit
```
doudou, line 1124: Actually, we should forbid calling `interrupt` - as we are not interrupting the proxied C++ component. Raise here, and handle read_only? in :stop itself.

lib/syskit/task_context.rb
doudou: > so the component will always be on ":PRE_OPERATIONAL".

No. It will be in RUNNING once the syskit that controls that component (a.k.a. has a non-readonly taskcontext on it) will start it.

In tests, you start the component manually within the test, just as I did in the code snippet I sent on the chat

lib/syskit/task_context.rb
caioaamaral: changed

lib/syskit/task_context.rb
caioaamaral: > In tests, you start the component manually within the test, just as I did in the code snippet I sent on the chat

okay `orocos_task.start(false) never emits`

```rb
Orocos.allow_blocking_calls do
                expect_execution { task.orocos_task.start(false) }
                    .to { emit task.start_event }
            end
```

The only way that I achieve to get the emission when using a "read_only" task was:

```rb
#Syskit::TaskContext
 event :start do |_context|
            info "starting #{self}"

            if read_only?
                event(:start).emit
                return
            end
            
[....]
```

 and then calling:

```
Orocos.allow_blocking_calls do
                expect_execution { task.start! }
                    .to { emit task.start_event }
end
```

But it's not really starting the component though...

caioaamaral: Unless I should not look into "task.orocos_task.rtt_state", but "task.started?" (from Roby::Task), is that so?

caioaamaral: An update:
Calling start(false) directly inside `allow_blocking_calls` does start the component

````rb
Orocos.allow_blocking_calls do
                task.orocos_task.start(false)
end
```

doudou: Just to be sure, you do need to also start the task, i.e.

```
expect_execution do
  task.orocos_task.start(false)
  task.start!
end.to { emit task.start_event }
```

If the event is not emitted, you need to find out why. The code responsible for this is in update_task_states.rb

caioaamaral: now I got it, done

test/test_readonly_task.rb
caioaamaral: done

lib/syskit/task_context.rb
caioaamaral: done

test/test_readonly_task.rb
```diff
@@ -0,0 +1,84 @@
...
+            it "never stop the component" do
+                task = @task
+
+                # here it's already stopped
+                expect_execution { task.stop! }
+                    .to { emit task.stop_event }
+
+                # starting the component {state will be :RUNNING}
+                Orocos.allow_blocking_calls { task.orocos_task.start(false) }
+                expect_execution { task.start! }
```
caioaamaral, line 74: Here the task fails to start due to: 
> Roby::TaskEventNotExecutable: start_event.emit on :0x5653221596d8<id:21>(conf: ["default"], orocos_name: stub3, read_only: true) which is not executable

For some reason, `task.executable?` is nil after explicitly calling `stop!`

doudou: This might be because you can't restart a Syskit task.

test/test_readonly_task.rb
```diff
@@ -0,0 +1,84 @@
...
+
+                expect_execution { task.start! }
+                    .to { emit task.start_event }
+            end
+
+            it "never stop the component" do
+                task = @task
+
+                # here it's already stopped
+                expect_execution { task.stop! }
```
doudou, line 69: How would that work ? You can't stop a task that is not running.

test/test_readonly_task.rb
```diff
@@ -0,0 +1,84 @@
...
+                expect_execution { task.start! }
+                    .to { emit task.start_event }
+
+                assert task.started?
+
+                expect_execution { task.start! }
+                    .to { emit task.start_event }
+            end
+
+            it "never stop the component" do
```
doudou, line 65: I don't understand what it is supposed to mean.

lib/syskit/task_context.rb
```diff
@@ -419,6 +442,25 @@ def commit_properties_if_needed(*args)
             end
         end
 
+        def start!
+            return super unless read_only?
+
+            event(:start).emit
+        end
```
doudou, line 449: Oh my.

Just return from the block in `event :start do |context| ... end`

This is very wrong.
- Calling an event can be done is more ways than calling `start!`
- it bypasses the checks
- it emits the event even if the remote task is not running

lib/syskit/task_context.rb
```diff
@@ -419,6 +442,25 @@ def commit_properties_if_needed(*args)
...
+
+        def stop!
+            return super unless read_only?
+
+            event(:stop).emit
+        end
```
doudou, line 455: Same. Return from the block in `event :stop do |context| ... end`

lib/syskit/task_context.rb
```diff
@@ -419,6 +442,25 @@ def commit_properties_if_needed(*args)
...
+
+        def check_emission_validity(event) # :nodoc:
+            return super unless read_only?
+
+            super unless running? && event.symbol == :start
+            super if started? && event.symbol == :stop
+        end
+
```
doudou, line 463: And no. Your job is to make sure the read only task behaves properly. Not to change the definition of "behaving properly"

lib/syskit/task_context.rb
```diff
@@ -1151,7 +1200,7 @@ def queue_last_chance_to_stop
 
         # Interrupts the execution of this task context
         event :stop do |_context|
-            interrupt!
+            interrupt! unless read_only?
```
doudou, line 1203: Here, you will need to emit stop (unlike for start)

The conflict comes from the meaning of 'stop' in this case. It may have two:
- the remote task stopped
- the Syskit app does not need to watch that remote task anymore

We don't have a way to separate the two here, but I think it does not matter in practice.

caioaamaral: If I emit stop, the `stop!` command will fail:

https://github.com/rock-core/tools-roby/blob/ff6744e5bd00e4e757df450ef0a62889915d0eaa/lib/roby/task_event_generator.rb#L257

test/test_readonly_task.rb
```diff
@@ -0,0 +1,61 @@
...
+                task = @task
+
+                Orocos.allow_blocking_calls do
+                    expect_execution { task.orocos_task.start(false) }
+                        .to { not_emit task.start_event }
+                end
+            end
+
+            it "emits start if it is already running and the component itself is started" do
+                task = @task
```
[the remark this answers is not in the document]
caioaamaral: done

test/test_readonly_task.rb
caioaamaral: I've updated this test case

lib/syskit/task_context.rb
doudou: > If stop is emitted, the stop! command will be rejected:

If stop is *already* emitted. Why would it be already emitted ?

test/test_readonly_task.rb
```diff
@@ -0,0 +1,71 @@
...
+            end
+
+            it "does not emit interrupted if the task is stopped" do
+                task = @task
+                Orocos.allow_blocking_calls { task.orocos_task.start(false) }
+                execute { task.start! }
+
+                expect_execution { task.stop! }
+                    .to { not_emit task.interrupt_event }
+            end
```
doudou, line 68: Need a test that checks what happens when the task is running but the component has been stopped.

doudou: (I think it should just emit stop)

doudou (commented):
I think we're almost there !

lib/syskit/task_context.rb
caioaamaral: > If stop is _already_ emitted. Why would it be already emitted ?

Sorry, I didn't get your point

caioaamaral: More  specifically this:
```rb
            it "does not emit interrupted if the task is stopped" do
                task = @task
                Orocos.allow_blocking_calls { task.orocos_task.start(false) }
                execute { task.start! }

                expect_execution { task.stop! }
                    .to { emit task.stop_event }
            end
```

causes: 

"Roby::CommandRejected: stop! called by [] but the task has finished. Task has been terminated by []."

Because when `task,stop!` is called, if it emits stops, the command itself will be rejected due to the way "check_call_validity" is written

caioaamaral: nevermind, my mistake

test/test_readonly_task.rb
```diff
@@ -0,0 +1,84 @@
...
+
+            it "does not emit start if started while the component is not running" do
+                task = @task
+
+                expect_execution do
+                    Orocos.allow_blocking_calls { task.orocos_task.start(false) }
+                end.to { not_emit task.start_event }
+            end
+
+            it "emits start if it is already running and the component itself is started" do
```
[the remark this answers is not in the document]
caioaamaral: changed

test/test_readonly_task.rb
```diff
@@ -0,0 +1,84 @@
...
+            it "emits start when it is started while the component is running" do
+                task = @task
+
+                expect_execution do
+                    Orocos.allow_blocking_calls { task.orocos_task.start(false) }
+                    task.start!
+                end.to { emit task.start_event }
+            end
+
+            it "does not emit start if started while the component is not running" do
```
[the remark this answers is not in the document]
caioaamaral: changed

test/test_readonly_task.rb
```diff
@@ -0,0 +1,84 @@
...
+                assert_raises(InvalidReadOnlyOperation) do
+                    task.properties.p = 2.0
+                end
+            end
+
+            it "emits start when it is started while the component is running" do
+                task = @task
+
+                expect_execution do
+                    Orocos.allow_blocking_calls { task.orocos_task.start(false) }
```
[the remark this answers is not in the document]
caioaamaral: changed

test/test_readonly_task.rb
caioaamaral: done

test/test_readonly_task.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
```
doudou, line 1: Sorry.... Just realized that this was a new file.

Test files have a 1:1 relationship with files in lib/. Keep that relationship.

In other words, these tests should be in `test_task_context.rb`. Just put them in a describe block to scope it

```
describe "read_only" do
end
```

doudou (approved):

caioaamaral:
Updated with last changes from master, lets wait for the ci

PR #322, opened 2022-05-17
feat: allow passing a service expressed as an InstanceRequirements in IR#select_service

doudou:
InstanceRequirements is the go-to object to represent ready-to-instanciate models (with e.g. arguments). This fixes a use-case, where one would need to inject a specific bound service, while also setting arguments.

### DIFF
--- a/test/test_instance_requirements.rb
+++ b/test/test_instance_requirements.rb
@@ -363,7 +363,10 @@ describe Syskit::InstanceRequirements do
             task_m = Syskit::TaskContext.new_submodel { provides Syskit::DataService.new_submodel, as: "srv" }
             assert_raises(ArgumentError) { req.select_service(task_m.srv_srv) }
         end
-        it "accepts selecting services from placeholder tasks if the set of models in the task matches the set of models in the instance requirements" do
+
+        it "accepts selecting services from placeholder tasks "\
+           "if the set of models in the task matches the set of models "\
+           "in the instance requirements" do
             srv_m  = Syskit::DataService.new_submodel
             task_m = srv_m.placeholder_model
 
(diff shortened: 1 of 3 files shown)


jhonasiv (approved):

test/test_instance_requirements.rb
```diff
@@ -374,6 +377,22 @@
...
+            srv_m  = Syskit::DataService.new_submodel
+            task_m = Syskit::TaskContext.new_submodel do
+                argument :key
+                provides srv_m, as: "test"
+            end
+
+            req = Syskit::InstanceRequirements.new([task_m])
+            srv = task_m.with_arguments(key: 10).test_srv
+            srv.name = "name" # closer to the original bug, which was a Definition
+            req_srv = req.select_service(srv)
```
caioaamaral, line 391: is srv `kind_of?(InstanceRequirements)`?

test/test_instance_requirements.rb
```diff
@@ -374,6 +377,22 @@
...
+        it "accepts a service selection expressed as an InstanceRequirements" do
+            srv_m  = Syskit::DataService.new_submodel
+            task_m = Syskit::TaskContext.new_submodel do
+                argument :key
+                provides srv_m, as: "test"
+            end
+
+            req = Syskit::InstanceRequirements.new([task_m])
+            srv = task_m.with_arguments(key: 10).test_srv
+            srv.name = "name" # closer to the original bug, which was a Definition
```
caioaamaral, line 390: what was the original bug? (is this a bugfix?)

caioaamaral (approved):

caioaamaral (approved):

test/test_instance_requirements.rb
caioaamaral: as I see it is a srv_m

doudou: It is. `task_m.test_srv` would be a `Models::BoundDataService`, but the `with_arguments` requires a `InstanceRequirements`

Generally speaking, both are interchangeable .... just not in this particular case.

test/test_instance_requirements.rb
doudou: Well ... it can be also seen as "something unsupported is now supported" ...

Was injecting in a profile something that looks exactly like `task_m.with_arguments(key: 10).test_srv` and it started to fail.

PR #323, opened 2022-05-19
fix(net_gen): do not start a logger we do not use

doudou:
This allows a (fragile but working) sharing of deployments across
Syskit instance, by disabling logging in one of the two instances.

### DIFF
--- a/lib/syskit/network_generation/logger.rb
+++ b/lib/syskit/network_generation/logger.rb
@@ -130,7 +130,6 @@ module Syskit
                     size: Syskit.conf.logs.default_logging_buffer_size
                 ]
 
-                seen_loggers = Set.new
                 engine.deployment_tasks.each do |deployment|
                     next unless deployment.plan
 
@@ -165,12 +164,10 @@ module Syskit
                     # Disconnect current log connections, we're going to
                     # reestablish the ones we want later on. We leave other
                     # connections as-is
-                    unless seen_loggers.include?(logger_task)
-                        dataflow = work_plan.task_relation_graph_for(Flows::DataFlow)
-                        deployment.each_executed_task do |t|
-                            if engine.deployed_tasks.include?(t)
-                                dataflow.remove_relation(t, logger_task)
-                            end
+                    dataflow = work_plan.task_relation_graph_for(Flows::DataFlow)
+                    deployment.each_executed_task do |t|
+                        if engine.deployed_tasks.include?(t)
+                            dataflow.remove_relation(t, logger_task)
                         end
                     end
 
--- a/lib/syskit/network_generation/logger.rb
+++ b/lib/syskit/network_generation/logger.rb
@@ -130,6 +130,7 @@ module Syskit
                     size: Syskit.conf.logs.default_logging_buffer_size
                 ]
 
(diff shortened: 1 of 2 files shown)


lib/syskit/network_generation/logger.rb
```diff
@@ -224,18 +224,9 @@ def self.add_logging_to_network(engine, work_plan)
...
-                end
-
-                # Mark as permanent any currently running logger
-                work_plan.find_tasks(logger_model)
-                         .not_finished
-                         .to_a.each do |t|
-                    work_plan.add_permanent_task(t)
+                required_loggers.each do |task|
+                    task.arguments[:conf] = ["default"] unless task.arguments[:conf]
+                    work_plan.add_permanent_task(task)
```
caioaamaral, line 229: Don't you have to check if "task" is finished anymore?

lib/syskit/network_generation/logger.rb
```diff
@@ -165,17 +165,17 @@ def self.add_logging_to_network(engine, work_plan)
...
+                    dataflow = work_plan.task_relation_graph_for(Flows::DataFlow)
+                    deployment.each_executed_task do |t|
+                        if engine.deployed_tasks.include?(t)
+                            dataflow.remove_relation(t, logger_task)
                         end
                     end
 
                     next if required_logging_ports.empty?
 
+                    required_loggers << logger_task
```
caioaamaral, line 177: so to disable logger in one of the deployment instances (when the deployment is shared), we just need to make `deployment.log_port?` false for every port its executed tasks have?

test/network_generation/test_logger.rb
```diff
@@ -115,6 +133,46 @@
...
+            logger = deployment.task "deployment_Logger"
+            deployment2.logger_task = logger
+
+            Syskit::NetworkGeneration::LoggerConfigurationSupport
+                .add_logging_to_network(syskit_engine, plan)
+
+            flexmock(deployment).should_receive(log_port?: false)
+            flexmock(deployment2).should_receive(log_port?: false)
+            Syskit::NetworkGeneration::LoggerConfigurationSupport
+                .add_logging_to_network(syskit_engine, plan)
```
caioaamaral, line 153: why do you have to call `#add_logging_to_network` twice?

caioaamaral (approved):

jhonasiv (approved):

lib/syskit/network_generation/logger.rb
doudou: Since I'm iterating over required_loggers, I am essentially assuming that we aren't relying on stopped loggers.

Which might or might not be the case... Not sure there is a test.

lib/syskit/network_generation/logger.rb
doudou: That is the nexus for the decision. There is however already a system for that, described here: https://www.rock-robotics.org/rock-and-syskit/log_management/runtime.html

test/network_generation/test_logger.rb
doudou: I'm essentially using the first one to setup connections, that are then cleaned up by the second.

PR #324, opened 2022-05-19
fix(cli): disable development mode in `syskit ide`

doudou:
The development mode turns on features that are specific to a running
Syskit instance, and are useless (and even could be harmful) in the IDE.

I discovered this because after I disabled the rebuilding of orogen
models, the IDE started reporting exceptions related to configuration
loading. At runtime, it was discovering configuration changes and tried
to reload the models using the (now empty) rebuilt models from remote.
Which obviously fails because they had no properties.


caioaamaral (approved):
So the `development mode` is completly unusable by now?

doudou:
No. It's not meant to be executed in the syskit IDE.

It works just fine where it is supposed to run, that is the main Syskit instance. This is what enables configuration file change detection for instance.

PR #325, opened 2022-05-24
feat: add read_only attribute to configured deployment

jhonasiv:
Depends on:
- [ ] https://github.com/rock-core/tools-orogen_syskit_tests/pull/3

On top of https://github.com/rock-core/tools-syskit/pull/321

### DIFF
--- a/lib/syskit/deployment.rb
+++ b/lib/syskit/deployment.rb
@@ -32,6 +32,7 @@ module Syskit
         argument :spawn_options, default: nil
         argument :ready_polling_period, default: 0.1
         argument :logger_task, default: nil
+        argument :read_only, default: []
 
         # The underlying process object
         attr_reader :orocos_process
@@ -210,7 +211,12 @@ module Syskit
                       "expected #{base_syskit_task_model} or one of its subclasses"
             end
 
-            plan.add(task = syskit_task_model.new(orocos_name: mapped_name))
+            is_read_only_model = read_only.any? do |m|
+                mapped_name.include?(m)
+            end
+            plan.add(task = syskit_task_model
+                            .new(orocos_name: mapped_name, read_only: is_read_only_model)
+            )
             task.executed_by self
             if scheduler_task
                 task.depends_on scheduler_task, role: "scheduler"
--- a/lib/syskit/deployment.rb
+++ b/lib/syskit/deployment.rb
@@ -211,11 +211,11 @@ module Syskit
                       "expected #{base_syskit_task_model} or one of its subclasses"
             end
 
-            is_read_only_model = read_only.to_a.any? do |m|
-                mapped_name.include?(m)
-            end
+            # Read only might be nil in some situations, that's why it's necessary to
+            # ensure this operation is done in an Array
+            is_read_only_task = read_only.to_a.include?(mapped_name)
             plan.add(task = syskit_task_model
-                            .new(orocos_name: mapped_name, read_only: is_read_only_model))
+                            .new(orocos_name: mapped_name, read_only: is_read_only_task))
             task.executed_by self
             if scheduler_task
                 task.depends_on scheduler_task, role: "scheduler"
--- a/lib/syskit/deployment.rb
+++ b/lib/syskit/deployment.rb
@@ -32,7 +32,8 @@ module Syskit
         argument :spawn_options, default: nil
         argument :ready_polling_period, default: 0.1
         argument :logger_task, default: nil
-        argument :read_only, default: []
+        argument :logger_name, default: nil
+        argument :read_only, default: nil
 
         # The underlying process object
         attr_reader :orocos_process
@@ -52,6 +53,7 @@ module Syskit
             @has_quarantines = false
             @quit_ready_event_monitor = Concurrent::Event.new
             @remote_task_handles = {}
+            self.read_only = [] unless read_only
             self.spawn_options = {} unless spawn_options
             self.name_mappings = {} unless name_mappings
             model.each_default_name_mapping do |k, v|
@@ -211,9 +213,7 @@ module Syskit
                       "expected #{base_syskit_task_model} or one of its subclasses"
             end
 
-            # Read only might be nil in some situations, that's why it's necessary to
-            # ensure this operation is done in an Array
-            is_read_only_task = read_only.to_a.include?(mapped_name)
+            is_read_only_task = read_only.include?(mapped_name)
             plan.add(task = syskit_task_model
                             .new(orocos_name: mapped_name, read_only: is_read_only_task))
             task.executed_by self
@@ -356,6 +356,10 @@ module Syskit
             process_server_config.log_dir
         end
 
+        def logger_name
+            arguments[:logger_name] || ("#{process_name}_Logger" if process_name)
+        end
+
         # Returns this deployment's logger
         #
         # @return [TaskContext,nil] either the logging task, or nil if this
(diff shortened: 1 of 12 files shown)


doudou:
Looks good, but please add some tests.

lib/syskit/network_generation/engine.rb
```diff
@@ -537,7 +537,8 @@ def reconfigure_tasks_on_static_port_modification(deployed_tasks)
                     work_plan
                     .find_tasks(Syskit::TaskContext).not_finished.not_finishing
                     .find_all do |t|
-                        deployed_tasks.include?(t) && (t.setting_up? || t.setup?)
+                        deployed_tasks.include?(t) &&
+                        (t.setting_up? || (t.setup? && !t.read_only?))
```
jhonasiv, line 541: @doudou I'm not sure why this was needed, but it solved an error that was occuring. It stated that in line 545, there was no `transaction.modifies_static_ports?` for `nil` class. Any idea if this is the correct solve?

lib/syskit/models/configured_deployment.rb
```diff
@@ -18,10 +18,14 @@ class ConfiguredDeployment
             # @return [Hash] the name mappings, e.g. the mapping from a task
             #   name in {#model} to the name this task should have while running
             attr_reader :name_mappings
+            # @return [Array] the tasks related to this deployment supposed to be set as
+            #   read only
+            attr_reader :read_only
```
doudou, line 23: This list should really be validated early. Possibly, ideally, at the creation of `ConfiguredDeployment`

That is, `ConfiguredDeployment` should check that all task names in `read_only` are actual names of the deployment.

Otherwise, it will fail at runtime, which is really something we want to avoid as much as possible.

lib/syskit/models/configured_deployment.rb
```diff
@@ -18,10 +18,14 @@ class ConfiguredDeployment
...
 
+            # rubocop:disable Metrics/ParameterLists
             def initialize(
                 process_server_name, model, name_mappings = {},
-                process_name = model.name, spawn_options = {}
+                process_name = model.name, spawn_options = {}, read_only = []
```
doudou, line 28: I think it would be better having `read_only` as a keyword argument (same for the other methods you modified in e.g. the tests).

lib/syskit/models/deployment_group.rb
```diff
@@ -459,10 +464,11 @@ def use_deployment(
                         process_managers.default_run_options(model)
                     )
 
+                    read_only_model = read_only ? [name] : []
```
doudou, line 467: This is wrong. `read_only` is meant to be a list of deployed task names, `name` is ... I don't know what actually.

Is there a test about this ?

lib/syskit/network_generation/engine.rb
```diff
@@ -537,7 +537,8 @@ def reconfigure_tasks_on_static_port_modification(deployed_tasks)
                     work_plan
                     .find_tasks(Syskit::TaskContext).not_finished.not_finishing
```
doudou, line 538: You should reject explicitly all read-only tasks here. This involves configuring, and that's something we don't do for them.

A.k.a. add `.find_all { |t| !t.read_only? }` before the last `find_all`

lib/syskit/test/stubs.rb
```diff
@@ -284,8 +284,8 @@ def handle_abstract_component_model_in_requirements(model)
             #
             # Computes a configured deployment suitable to deploy the given task model
             def stub_configured_deployment(
-                task_model = nil, task_name = default_stub_name, remote_task: false,
-                &block
+                task_model = nil, task_name = default_stub_name, read_only = [],
```
doudou, line 287: make `read_only` a keyword argument here.

lib/syskit/deployment.rb
```diff
@@ -210,7 +211,11 @@ def create_deployed_task(
                       "expected #{base_syskit_task_model} or one of its subclasses"
             end
 
-            plan.add(task = syskit_task_model.new(orocos_name: mapped_name))
+            is_read_only_model = read_only.to_a.any? do |m|
+                mapped_name.include?(m)
+            end
```
doudou, line 216: `mapped_name` is just a name. This is `read_only.include?(mapped_name)`

lib/syskit/models/deployment_group.rb
```diff
@@ -459,10 +464,11 @@ def use_deployment(
                         process_managers.default_run_options(model)
                     )
 
+                    read_only_deployment = read_only ? [deployment_name] : []
```
doudou, line 467: Still wrong.

A deployment may have more than one task. Turns out that *if* you are using the default deployment `OroGen.project.Task => "name"` the two are the same, but it is not true in general.

`name_mappings` is complete AFAIK. That is, all values in `name_mappings` are all the tasks in the deployment. So, you can use `name_mappings.keys`

However, I would also let the caller pass specific names, i.e. you need to check whether `read_only` is an array before you set it to that.

lib/syskit/deployment.rb
jhonasiv: The idea behind using it like this is to allow setting a deployment as read_only using only a substring. 

Let's say there is a gazebo model named `vessel` on the world, you can set all of the model's links, plugins, etc as read_only by just setting read_only: ["vessel"]. To set everything of the world as read_only, use ["gazebo"], for example. This way you dont need to express all the links, joints and plugins as read_only.

Using `read_only.include?(mapped_name)` would return true only if the exact string is present on the array.

lib/syskit/models/deployment_group.rb
jhonasiv: Yeah, the test is on `test_deployment_group`.

lib/syskit/deployment.rb
jhonasiv: I am thinking this can be improved by adding another argument like `read_only_exact_match: true`. This way the user would have control whether to match the exact word or the substring.

lib/syskit/models/deployment_group.rb
doudou: This is not that. A deployment can have multiple tasks, that's what you are not handling here. You are already dealing with the `use_` resolving multiple times...

lib/syskit/deployment.rb
```diff
@@ -210,7 +211,11 @@ def create_deployed_task(
                       "expected #{base_syskit_task_model} or one of its subclasses"
             end
 
-            plan.add(task = syskit_task_model.new(orocos_name: mapped_name))
+            # Read only might be nil in some situations, that's why it's necessary to
+            # ensure this operation is done in an Array
+            is_read_only_task = read_only.to_a.include?(mapped_name)
```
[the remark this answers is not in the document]
jhonasiv: Some deployments are starting read_only as nil, ignoring the default value, I didnt manage to find out why

lib/syskit/models/configured_deployment.rb
```diff
@@ -34,7 +39,32 @@ def initialize(
...
+
+            # @api private
+            #
+            # Resolves the read_only argument into an Array of task names to be set as
+            # read_only.
+            #
+            # @return [Array<String>]
+            def resolve_read_only(read_only, mappings)
+                non_logger_names = mappings.values.reject { |name| /_Logger$/ === name }
+                return non_logger_names if read_only == true
```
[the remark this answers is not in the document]
jhonasiv: To support read only loggers we should alter its configuration process, which we dont rn. Do you think it's better to do it in this PR? If not, any suggestions on making it clearer to the user that loggers cant be read only at the moment?

PR #331, opened 2022-08-12
fix: do not allow read-only if 'Roby.app.single'

caioaamaral:
During tests, read-only tasks would fail to initialize, leading to strange errors during test case's Teardown

### DIFF
--- a/lib/syskit/task_context.rb
+++ b/lib/syskit/task_context.rb
@@ -171,7 +171,7 @@ module Syskit
         #
         # By default, its false
         def read_only?
-            read_only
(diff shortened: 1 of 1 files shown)


jhonasiv (approved):

doudou:
@caioaamaral can you explain why the strange errors, and while single? solves it, or should I ?

caioaamaral:
> @caioaamaral can you explain why the strange errors, and while single? solves it, or should I ?

I would appreciate having an explanation on your words

doudou (approved):

PR #333, opened 2022-09-15
fix: update syskit start all deployments method 

jhonasiv:
It needed a `each_configured_deployment` method from `Syskit.conf` which doesnt exist anymore.

### DIFF
--- a/lib/syskit/models/deployment_group.rb
+++ b/lib/syskit/models/deployment_group.rb
@@ -225,6 +225,9 @@ module Syskit
 
             # Enumerates all the deployments registered on self
             #
+            # @param [String] on the process server names to look for. If empty, all
+            #   process server names will be considered.
+            # @param [String] except_on the process server names to be ignored.
             # @yieldparam [ConfiguredDeployment]
             def each_configured_deployment(on: [], except_on: [])
                 return enum_for(__method__, on: on, except_on: except_on) unless block_given?
--- a/lib/syskit/models/deployment_group.rb
+++ b/lib/syskit/models/deployment_group.rb
@@ -226,15 +226,13 @@ module Syskit
             # Enumerates all the deployments registered on self
             #
             # @yieldparam [ConfiguredDeployment]
-            def each_configured_deployment(on: nil, except_on: "")
-                return enum_for(__method__) unless block_given?
+            def each_configured_deployment(on: [], except_on: [])
+                return enum_for(__method__, on: on, except_on: except_on) unless block_given?
 
                 deployments.each_value do |set|
                     set.each do |c|
-                        if !on.nil? && c.process_server_name == on || \
-                           on.nil? && c.process_server_name != except_on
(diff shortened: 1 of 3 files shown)


lib/syskit/roby_app/plugin.rb
```diff
@@ -727,16 +727,27 @@ def using_ros_launcher(name, options = {})
...
                                            .not_finished
                                            .find_all(&:reusable?)
                                            .map(&:process_name).to_set
 
-                Syskit.conf.each_configured_deployment(on: on, except_on: except_on) do |configured_deployment|
-                    next if existing_deployments.include?(configured_deployment.process_name)
+                deployment_group.each_configured_deployment(
+                          on: on,
+                          except_on: except_on
+                ) do |configured_deployment|
```
jhonasiv, line 747: I tried recreating the interface as it was in the previous version, but I am not sure if it makes sense to filter for the process server name on the `deployment_group` side, can a deployment group have more than one process server name? If not, filtering for `on` and `except_on` could be done inside this method, leaving `deployment_group.each_configured_deployment` unchanged.

lib/syskit/models/deployment_group.rb
```diff
@@ -225,12 +225,18 @@ def register_configured_deployment(configured_deployment)
...
-            def each_configured_deployment
-                return enum_for(__method__) unless block_given?
+            def each_configured_deployment(on: [], except_on: [])
+                return enum_for(__method__, on: on, except_on: except_on) unless block_given?
 
                 deployments.each_value do |set|
-                    set.each { |c| yield(c) }
+                    set.each do |c|
+                        yield(c) if !on.empty? && on.include?(c.process_server_name)
+                        yield(c) if on.empty? && !except_on.include?(c.process_server_name)
```
doudou, line 238: This hides that the two conditions are mutually exclusive ... which is weird. Maybe

```
if on.empty?
   yield(c) if except_on.include?(c.process_server_name)
else
   yield(c) if on.include?(c.process_server_name)
end
```

lib/syskit/roby_app/plugin.rb
doudou: It does make sense generally speaking ... but also find it weird. It feels more flexible and not worse to do ad-hoc filtering in each method that need it, instead of having an inherently limited interface in each_configured_deployment.

lib/syskit/roby_app/plugin.rb
```diff
@@ -727,16 +727,34 @@ def using_ros_launcher(name, options = {})
...
+                    next if existing_deployments.include?(
+                        configured_deployment.process_name
+                    )
+
+                    if on.empty?
+                        next if except_on.include?(
+                            configured_deployment.process_server_name
+                        )
+                    else
+                        next if !on.include?(
```
doudou, line 754: `next unless`

doudou (approved):

PR #334, opened 2022-09-20
feat(app): optionally call kill_all on newly connected process servers

doudou:
This should allow to recover from Syskit crashes (provided that a watchdog kills the Syskit process). It does not apply to local process servers, as they are always started fresh. But local process servers (and locally started components) can be terminated by e.g. systemd along with syskit itself. This is meant to deal with the remote ones.


doudou:
Actually forgot that commit in the last PR ...

jhonasiv (approved):

caioaamaral (approved):

PR #336, opened 2022-10-24
chore(net_gen): remove extra ')' from a warn message

caioaamaral:

PR #340, opened 2022-11-28
fix: make log transfer operational

doudou:
Depends on:
 - [ ] https://github.com/tidewise/bundles-wetpaint/pull/257

These PR adds rate limiting, which was a needed feature. The rest is to make the setup of it easier, and have a way to transfer data even for localhost (great to test, and will be useful since we want to have a dedicated log server)


jhonasiv (approved):

lib/syskit/network_generation/logger.rb
```diff
@@ -129,7 +129,7 @@ def self.add_logging_to_network(engine, work_plan)
                     unless (logger_task = deployment.logger_task)
                         warn "deployment #{deployment.process_name} has no usable "\
                              "logger (default logger name would be "\
-                             "#{deployment.process_name}_Logger))"
+                             "#{deployment.process_name}_Logger)"
```
caioaamaral, line 132: I had fixed that, but my target was on master instead of 'transition-to-runkit' https://github.com/rock-core/tools-syskit/pull/336

every PR on tools-syskit (from our side) should target `transition-to-runkit`?

doudou: Until we get the runkit branch merged, yes, I'm afraid.

It should merge cleanly, so I don't think it's an issue.

caioaamaral (approved):

PR #342, opened 2022-12-07
fix: use log_dir as default log transfer target dir

doudou:


doudou:
What is very strange is that the build for #340 should have caught this. I'm a bit lost.

doudou:
Got it. The issue is was in the bundle's usage of feature flags. They were initialized in `setup` and the log transfer has to be setup in `init`. My latest PR in bundles/wetpaint moved feature flags setup in `init` too, which triggered the bug.

PR #351, opened 2023-02-13
fix: interferences between a syskit with a readonly task and the remote syskit that handles it

jhonasiv:


jhonasiv:
@doudou buggering you to take a look at this

doudou:
@jhonasiv have a look at the cleanup I did and decide when to merge

PR #354, opened 2023-02-17
fix: properly handle action methods with required arguments in profile assertions

doudou:
The profile assertion action resolving methods try to instanciate method tasks, but was either (1) generating an error if the action method has missing required arguments *and* the action was toplevel or (2) generating an error claiming we could pass it to `exclude`
- but we actually could not because of a bug *and* it was not the right suggestion.

Fix both cases, add documentation and tests.

On the exclude bug: Syskit tests that are given actions are instanciating the actions to figure out all the syskit definitions that are involved. But the same code path was used for the `exclude` argument (instanciate and then convert back to an action model), even though the exclusion is not recursive (i.e. we only want the model associated with the excluded object, not whatever action is derived from it).

### DIFF
--- a/lib/syskit/test/profile_assertions.rb
+++ b/lib/syskit/test/profile_assertions.rb
@@ -31,6 +31,14 @@ module Syskit
                 end
             end
 
+            # @api private
+            #
+            # Find all planning tasks in a task hierarchy and return the resolved actions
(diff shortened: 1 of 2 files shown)


doudou:
Closed in favor of #354

PR #356, opened 2023-03-16
fix: forward explicitly set services

jhonasiv:


doudou:
Ping @wvmcastro

test/actions/test_profile.rb
```diff
@@ -226,6 +226,22 @@ module Actions
                                         .requirements.instanciate(plan)
                     assert_kind_of srv_task_m, task.test_child
                 end
+
+                it "forwards explicitly selected services" do
+                    srv_m = DataService.new_submodel
+                    task_m = TaskContext.new_submodel
+                    task_m.provides srv_m, as: "srv1"
+                    task_m.provides srv_m, as: "srv2"
```
wvmcastro, line 234: how providing the same service twice is relevant for this test?

doudou: It makes sure that the system does not fall back on the implicit service selection code path (since that code path would fail). The caller *must* explicitly select a service.

wvmcastro (approved):

PR #360, opened 2023-04-20
fix: apply example arguments to atomic actions in profile assertions

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/242

This allows to handle a lot more actions with arguments in bulk tests (e.g. assert_can_deploy) instead of having to pass them explicitly


doudou:
Ping

jhonasiv (approved):

PR #361, opened 2023-04-24
feat: generate action and robot test files that run assert_can_deploy

doudou:
On top of #360

### DIFF
--- a/lib/syskit/network_generation/dataflow_computation.rb
+++ b/lib/syskit/network_generation/dataflow_computation.rb
@@ -154,6 +154,20 @@ module Syskit
                 end
             end
 
+            # @api private
+            #
+            # For testing purposes only
+            def changed?
+                @changed
+            end
+
+            # @api private
+            #
+            # For testing purposes only
+            def reset_changed
+                @changed = false
+            end
+
             def reset(tasks = [])
                 @result = Hash.new { |h, k| h[k] = {} }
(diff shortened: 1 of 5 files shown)


lib/syskit/cli/gen/syskit_app/config/robots/robot.rb.erb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
```
jhonasiv: No Robot.config?

jhonasiv (approved):

lib/syskit/cli/gen/syskit_app/config/robots/robot.rb.erb
doudou: Robot.config is replaced by Robot.setup. `config` was a confusing name from Syskit's perspective. Check out the comment.

If you are confused by this, maybe we should find a way to make it more obvious. Do you have a suggestion ?

jhonasiv: I dont have a problem with the name. It's more that it wasnt clear what were all the possible steps, and since some of the code I've seen used some of the old names (i.e. Robot.config) I wasnt sure what was the exact order of execution of the stages. Maybe add a deprecated warning on the replaced names?

PR #362, opened 2023-04-26
fix: change setup for read only tasks

jhonasiv:
Instead of always considering that a read_only task setup is done, we will consider that it is done whenever its not-read only counterpart is running. Also, read_only tasks do nothing on perform_setup.

### DIFF
--- a/lib/syskit/task_context.rb
+++ b/lib/syskit/task_context.rb
@@ -694,20 +694,34 @@ module Syskit
                 return
             end
 
-            return orocos_task.runtime_state?(state) if read_only?
-
-            configurable_state =
-                CONFIGURABLE_RTT_STATES.include?(state) ||
-                orocos_task.exception_state?(state)
-            if configurable_state
-                true
+            if read_only?
+                read_only_task_verify_configurable_state(state)
             else
-                debug do
-                    "#{self} not ready for setup: in state #{state}, "\
-                    "expected STOPPED, PRE_OPERATIONAL or an exception state"
-                end
-                false
+                normal_task_verify_configurable_state(state)
             end
+        end
+
+        def read_only_task_verify_configurable_state(state)
+            return true if orocos_task.runtime_state?(state)
+
+            execution_engine.scheduler.report_holdoff(
+                "read-only task #{self} not ready, task is not currently running", self
+            )
+            false
+        end
+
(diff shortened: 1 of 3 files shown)


lib/syskit/task_context.rb
```diff
@@ -672,6 +672,8 @@ def read_current_state
         # Returns true if this component needs to be setup by calling the
         # #setup method, or if it can be used as-is
         def ready_for_setup?(state = nil)
+            return running? if read_only?
```
doudou: This is wrong. `running?` refers to the Roby task. When we check if we should set it up, `running?` will always be false.

Look later in this very method. There is a check for the component state to be an exception state, STOPPED or PRE_OPERATIONAL. You would need to check whether the state is a running state in the case of read-only

And add tests ...

lib/syskit/test/network_manipulation.rb
```diff
@@ -545,6 +545,8 @@ def syskit_configure(
                     current_pending = pending.size
                     has_missing_states = false
                     pending.delete_if do |t|
+                        next true if t.read_only?
+
```
doudou: This *must* not be needed.

test/test_task_context.rb
```diff
@@ -700,6 +700,34 @@ def push_task_state(state)
...
+            before do
+                task = syskit_stub_and_deploy("Task") {}
+                syskit_start_execution_agents(task)
+                @task = flexmock(task)
+            end
+
+            it "does nothing when the task is read_only" do
+                task.should_receive(:read_only?).and_return(true)
+                task.should_receive(:prepare_for_setup).never
+                flexmock(task.orocos_task).should_receive(:configure).never
```
doudou: You still need to actually do something ...

test/test_task_context.rb
```diff
@@ -2336,6 +2333,16 @@ def assert_process_events_does_not_block
                 end
             end
 
+            it "does not perform setup on configuration" do
+                task = deployment.task("test")
+                flexmock(task).should_receive(:perform_setup).once
```
doudou: Mocking a method makes it not being called *at all*. I.e. this mock ensures that the next one is never called.

You have to add `.pass_thru` to let the call go through.

jhonasiv:
Ping @doudou

doudou (approved):

PR #379, opened 2023-09-19
fix: setup log rotation for runtime only

doudou:
This pull request fixes problems opening, with e.g. the syskit IDE, a Syskit application that sets up log transfer. In that
case, the built-in log transfer server was setup in `#setup`, that is was actually setup when opening `syskit ide`,
and that failed because we can't bind to the target address.

### DIFF
--- a/lib/syskit/roby_app/plugin.rb
+++ b/lib/syskit/roby_app/plugin.rb
@@ -261,10 +261,11 @@ module Syskit
                 @handler_ids = plug_engine_in_roby(app.execution_engine)
 
                 if Syskit.conf.log_rotation_period
-                    app.execution_engine.every(Syskit.conf.log_rotation_period) do
-                        app.syskit_log_perform_rotation_and_transfer
-                        app.syskit_log_transfer_poll_state
-                    end
+                    @log_rotation_poll_handler =
+                        app.execution_engine.every(Syskit.conf.log_rotation_period) do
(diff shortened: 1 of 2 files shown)


doudou:
Ping @higor-melo

lib/syskit/roby_app/plugin.rb
```diff
@@ -282,6 +281,9 @@ def self.shutdown(app)
                     unplug_engine_from_roby(@handler_ids.values, app.execution_engine)
                     @handler_ids = nil
                 end
+
+                @log_rotation_poll_handler&.dispose
```
higor-melo, line 285: why do we still need to dispose of the `log_rotation_poll_handler` if we are shutting down the log transfer manager? and if we need to do it, can we move this to the `def syskit_log_transfer_shutdown`?

doudou: These are two different features.

Log rotation locally creates new log files every X minutes. That is, for a ublox_gps deployment, it makes the logger create ublox_gps.0.log, ublox_gps.1.log, ublox_gps.2.log

Log transfer is used in distributed systems (e.g. Tupan with the beagleboards). It makes the satellite systems copy their rotated logs to a central location to regroup them and ease management later (and avoid filling a BBB flash memory which is limited)

higor-melo: Got it

wvmcastro (approved):

PR #386, opened 2023-12-19
feat: add free space management to the archiver

rayssinha-soares:
Depends on:
- [ ] https://github.com/rock-core/package_set/pull/226

### DIFF
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -47,10 +47,10 @@ module Syskit
             # The method will check if there is enough space to save more log files
             # according to pre-established threshold.
             #
-            # @param [integer] free_space_low_limit: required free space threshold, at
-            #   which the archiver starts deleting the oldest log files
-            # @param [integer] free_space_delete_until: post-deletion free space, at which
-            #   the archiver stops deleting the oldest log files
+            # @param [integer] free_space_low_limit: required free space threshold in
+            #   bytes, at which the archiver starts deleting the oldest log files
+            # @param [integer] free_space_delete_until: post-deletion free space in bytes,
+            #   at which the archiver stops deleting the oldest log files
             def ensure_free_space(free_space_low_limit, free_space_delete_until)
                 if free_space_low_limit > free_space_delete_until
                     raise ArgumentError,
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -13,20 +13,12 @@ module Syskit
         # It depends on the syskit instance using log rotation
         class LogRuntimeArchive
             DEFAULT_MAX_ARCHIVE_SIZE = 10_000_000_000 # 10G
-            FREE_SPACE_LOW_LIMIT = 5_000_000_000 # 5 G
-            FREE_SPACE_FREED_LIMIT = 25_000_000_000 # 25 G
 
             def initialize(
                 root_dir, target_dir,
                 logger: LogRuntimeArchive.null_logger,
                 max_archive_size: DEFAULT_MAX_ARCHIVE_SIZE
             )
-                if FREE_SPACE_LOW_LIMIT > FREE_SPACE_FREED_LIMIT
-                    raise ArgumentError,
-                          "cannot erase files: freed limit is smaller than " \
-                          "low limit space."
-                end
-
                 @last_archive_index = {}
                 @logger = logger
                 @root_dir = root_dir
@@ -59,8 +51,13 @@ module Syskit
             #   which the archiver starts deleting the oldest log files
             # @param [integer] free_space_delete_until: post-deletion free space, at which
             #   the archiver stops deleting the oldest log files
-            def ensure_free_space(free_space_low_limit = FREE_SPACE_LOW_LIMIT,
-                free_space_delete_until = FREE_SPACE_FREED_LIMIT)
+            def ensure_free_space(free_space_low_limit, free_space_delete_until)
+                if free_space_low_limit > free_space_delete_until
+                    raise ArgumentError,
+                          "cannot erase files: freed limit is smaller than " \
+                          "low limit space."
+                end
+
                 stat = Sys::Filesystem.stat(@target_dir)
                 available_space = stat.bytes_free
 
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -64,15 +64,14 @@ module Syskit
                 stat = Sys::Filesystem.stat(@target_dir)
                 available_space = stat.bytes_free
 
-                return if available_space > free_space_delete_until
+                return if available_space > free_space_low_limit
 
                 until available_space >= free_space_delete_until
-                    files = @target_dir.each_child.select { |file| file.file? }
+                    files = @target_dir.each_child.select(&:file?)
                     break if files.empty?
 
+                    available_space += files.sort.first.size
                     files.sort.first.unlink
-                    stat = Sys::Filesystem.stat(@target_dir)
-                    available_space = stat.bytes_free
                 end
             end
 
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -59,21 +59,23 @@ module Syskit
             #   the archiver stops deleting the oldest log fil
             def ensure_free_space(free_space_low_limit: @free_space_low_limit,
                     free_space_delete_until: @free_space_delete_until)
+                return if free_space_low_limit >= free_space_delete_until
+
                 stat = Sys::Filesystem.stat(@target_dir)
                 free_space = stat.bytes_free
 
                 return if free_space > free_space_low_limit
 
-                candidates = each_file_from_path(path).to_a.sort
-                loop do |index|
-                    stat.delete(candidates[index])
+                candidates = Dir.entries(@target_dir).sort
+                until free_space >= free_space_delete_until
+                    candidates.shift
                     free_space = stat.bytes_free
-                    break if free_space >= free_space_delete_until
                 end
             end
 
+            # Returns the file size in bytes
             def self.size_of_file(path)
-                path.stat.size
+                File.stat(path).size
             end
 
             def process_dataset(child, full:)
@@ -98,7 +100,6 @@ module Syskit
                     use_existing = false
                 end
             end
-
 
             # Create or open an archive
             #
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -79,11 +79,6 @@ module Syskit
                 end
             end
 
-            # Returns the file size in bytes
-            def self.size_of_file(path)
-                File.stat(path).size
-            end
-
             def process_dataset(child, full:)
                 use_existing = true
                 loop do
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -23,6 +23,12 @@ module Syskit
                 free_space_low_limit: FREE_SPACE_LOW_LIMIT,
                 free_space_delete_until: FREE_SPACE_FREED_LIMIT
             )
+                if free_space_low_limit > free_space_delete_until
+                    raise ArgumentError,
+                          "cannot erase files: freed limit is smaller than " \
+                          "low limit space."
+                end
+
                 @last_archive_index = {}
                 @logger = logger
                 @root_dir = root_dir
@@ -59,8 +65,6 @@ module Syskit
             #   the archiver stops deleting the oldest log files
             def ensure_free_space(free_space_low_limit: @free_space_low_limit,
                 free_space_delete_until: @free_space_delete_until)
-                return if free_space_low_limit >= free_space_delete_until
-
                 stat = Sys::Filesystem.stat(@target_dir)
                 available_space = stat.bytes_free
 
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -48,7 +48,7 @@ module Syskit
                 end
             end
 
-            # Manages folder free space
+            # Manages folder available space
             #
             # The method will check if there is enough space to save more log files
             # according to pre-established threshold.
@@ -56,20 +56,23 @@ module Syskit
             # @param [integer] free_space_low_limit: required free space threshold, at
             #   which the archiver starts deleting the oldest log files
             # @param [integer] free_space_delete_until: post-deletion free space, at which
-            #   the archiver stops deleting the oldest log fil
+            #   the archiver stops deleting the oldest log files
             def ensure_free_space(free_space_low_limit: @free_space_low_limit,
                     free_space_delete_until: @free_space_delete_until)
                 return if free_space_low_limit >= free_space_delete_until
 
                 stat = Sys::Filesystem.stat(@target_dir)
-                free_space = stat.bytes_free
+                available_space = stat.bytes_free
 
-                return if free_space > free_space_low_limit
+                return if available_space > free_space_delete_until
 
-                candidates = Dir.entries(@target_dir).sort
-                until free_space >= free_space_delete_until
-                    candidates.shift
-                    free_space = stat.bytes_free
+                until available_space >= free_space_delete_until
+                    files = @target_dir.each_child.select { |file| file.file? }
+                    break if files.empty?
+
+                    files.sort.first.unlink
+                    stat = Sys::Filesystem.stat(@target_dir)
+                    available_space = stat.bytes_free
                 end
             end
 
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -19,11 +19,9 @@ module Syskit
             def initialize(
                 root_dir, target_dir,
                 logger: LogRuntimeArchive.null_logger,
-                max_archive_size: DEFAULT_MAX_ARCHIVE_SIZE,
-                free_space_low_limit: FREE_SPACE_LOW_LIMIT,
-                free_space_delete_until: FREE_SPACE_FREED_LIMIT
+                max_archive_size: DEFAULT_MAX_ARCHIVE_SIZE
             )
-                if free_space_low_limit > free_space_delete_until
+                if FREE_SPACE_LOW_LIMIT > FREE_SPACE_FREED_LIMIT
                     raise ArgumentError,
                           "cannot erase files: freed limit is smaller than " \
                           "low limit space."
@@ -34,8 +32,6 @@ module Syskit
                 @root_dir = root_dir
                 @target_dir = target_dir
                 @max_archive_size = max_archive_size
-                @free_space_low_limit = free_space_low_limit
-                @free_space_delete_until = free_space_delete_until
             end
 
             # Iterate over all datasets in a Roby log root folder and archive them
@@ -63,8 +59,8 @@ module Syskit
             #   which the archiver starts deleting the oldest log files
             # @param [integer] free_space_delete_until: post-deletion free space, at which
             #   the archiver stops deleting the oldest log files
-            def ensure_free_space(free_space_low_limit: @free_space_low_limit,
-                free_space_delete_until: @free_space_delete_until)
+            def ensure_free_space(free_space_low_limit = FREE_SPACE_LOW_LIMIT,
+                free_space_delete_until = FREE_SPACE_FREED_LIMIT)
                 stat = Sys::Filesystem.stat(@target_dir)
                 available_space = stat.bytes_free
 
@@ -72,10 +68,16 @@ module Syskit
 
                 until available_space >= free_space_delete_until
                     files = @target_dir.each_child.select(&:file?)
-                    break if files.empty?
+                    if files.empty?
+                        Roby.warn "Cannot erase files: the folder is empty but the "\
+                        "available space is smaller than the threshold."
+                        break
+                    end
 
-                    available_space += files.sort.first.size
-                    files.sort.first.unlink
+                    removed_file = files.min
+                    size_removed_file = removed_file.size
+                    removed_file.unlink
+                    available_space += size_removed_file
                 end
             end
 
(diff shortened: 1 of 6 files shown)


lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -41,6 +48,30 @@ def process_root_folder
...
+            #   the archiver stops deleting the oldest log fil
+            def ensure_free_space(free_space_low_limit: @free_space_low_limit,
+                    free_space_delete_until: @free_space_delete_until)
+                stat = Sys::Filesystem.stat(@target_dir)
+                free_space = stat.bytes_free
+
+                return if free_space > free_space_low_limit
+
+                candidates = each_file_from_path(path).to_a.sort
+                loop do |index|
```
doudou:
```
loop do
    ....
    break if <condition>
end
```

is a while. Ruby has both `while` and `until`. I would have personally preferred

```
until free_space >= free_space_delete_until
end
```

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -12,17 +13,23 @@ module CLI
...
+                free_space_low_limit: FREE_SPACE_LOW_LIMIT,
+                free_space_delete_until: FREE_SPACE_FREED_LIMIT
             )
                 @last_archive_index = {}
                 @logger = logger
                 @root_dir = root_dir
                 @target_dir = target_dir
                 @max_archive_size = max_archive_size
+                @free_space_low_limit = free_space_low_limit
+                @free_space_delete_until = free_space_delete_until
```
doudou: Validate that `free_space_low_limit > free_space_delete_until`

jhonasiv: You meant `free_space_low_limit` < `free_space_delete_until`, right? 

@rayssinha-soares you marked as resolved but I dont see where you are validating it

test/cli/test_log_runtime_archive.rb
```diff
@@ -510,6 +510,102 @@ def should_archive_dataset(dataset, archive_basename, full:)
                 end
             end
 
+            describe "#ensure_free_space" do
```
jhonasiv, line 513: I dont get what are the "files" supposed to be. Are they temporary files created in memory?

test/cli/test_log_runtime_archive.rb
```diff
@@ -510,6 +510,102 @@ def should_archive_dataset(dataset, archive_basename, full:)
...
+                        end
+                end
+
+                def compute_mocked_disk_size(total_disk_size)
+                    return total_disk_size if @mocked_files_sizes.empty?
+
+                    @mocked_files_sizes.each_index do |i|
+                        if !(@archive_dir / i.to_s).exist?
+                            total_disk_size -= @mocked_files_sizes[i]
+                            @deleted_files.append(i)
```
jhonasiv: You are taking all the files that are actually deleted, not the ones that are supposed to be deleted. Instead of computing the deleted files in here, you should just have a list of the ones that you expect to delete for each unit test.

jhonasiv (changes requested):

lib/syskit/cli/log_runtime_archive.rb
rayssinha-soares: I added this condition: [`return if free_space_low_limit >= free_space_delete_until`](https://github.com/rock-core/tools-syskit/pull/386/files#diff-27a24ddb639c1c67b459feffe67d0e7556ab8a2dba9de22165c32aef63d9fb4bR62)

test/cli/test_log_runtime_archive.rb
rayssinha-soares: > Are they temporary files created in memory?

yes, the `@archive_dir = make_tmppath` in line 515 makes a temporary directory that is removed on test teardown
image

lib/syskit/cli/log_runtime_archive.rb
jhonasiv: That's not validation, a validation statement would throw in case of invalid values.

About the low_limit/delete_until relation, what do you guys have in mind? I thought if low_limit < delete_until, it would be a scenario like (which is examplified by the defaults of both of these variables):
```
low_limit = 10G
delete_until = 15G

if available < low_limit
     delete files until there is 15G available
end
```

But if low_limit > delete until, then once it reaches the low limit, it would stay in a cycle deleting files until some outside actor frees enough disk to reach > low limit again.

test/cli/test_log_runtime_archive.rb
jhonasiv: Yeah, but thats a directory, are you mocking real files inside this directory? I couldnt figure out while reading the tests

jhonasiv: Just now I saw that you do write the files in the tmp_dir, got it.

test/cli/test_log_runtime_archive.rb
rayssinha-soares: > You are taking all the files that are actually deleted, not the ones that are supposed to be deleted.

Actually that's the main idea of this function... to mock the avaiable space in folder every time a file is deleted. So I do need to compare with files that are deleted.

>Instead of computing the deleted files in here, you should just have a list of the ones that you expect to delete for each unit test.

 I did this in `assert_deleted_files` function.

jhonasiv: Except that you call `assert_deleted_files` passing the @deleted_files variable, which is the list of files that were deleted, not the list of files that you *expect* to be deleted.

See the dataflow of the variable, it is filled with all files that no longer exist by checking the `@archive_dir / i` path. Then, on assert_deleted_files, you check if the files that were deleted still exist (spoiler: they dont because you checked that to know if they were deleted), or that the files that werent deleted are still there.

lib/syskit/cli/log_runtime_archive.rb
rayssinha-soares: done

test/cli/test_log_runtime_archive.rb
rayssinha-soares: fixed

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -41,6 +54,36 @@ def process_root_folder
...
+
+                    available_space += files.sort.first.size
+                    files.sort.first.unlink
+                end
+            end
+
+            # Returns the file size in bytes
+            def self.size_of_file(path)
+                File.stat(path).size
+            end
```
jhonasiv: You dont really use this here, do you have any plans of using it in the future?

doudou: No. I created it in the first version, but @rayssinha-soares went another way that does not require this anymore. Need to be removed @rayssinha-soares

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -12,17 +13,29 @@ module CLI
         # It depends on the syskit instance using log rotation
         class LogRuntimeArchive
             DEFAULT_MAX_ARCHIVE_SIZE = 10_000_000_000 # 10G
+            FREE_SPACE_LOW_LIMIT = 1_000_000_000 # 1 G
+            FREE_SPACE_FREED_LIMIT = 10_000_000_000 # 10 G
```
doudou: The default archive size is 10G. I feel that these low/freed limits are on the low side.

The objective is that the free-space manager lets the archiver do its job. In practice, this means that within the default 10 min period, we need to generate less than freed_limit compressed data. We already do (we do ~10G/hour)

Let's baseline 20G/hour. That's 3G per 10 min. I think the low limit should be 5G and the freed limit 25G to let some room to breathe.

jhonasiv (approved):

lib/syskit/scripts/log_runtime_archive.rb
```diff
@@ -18,12 +18,21 @@ def self.exit_on_failure?
...
+    option :free_space_delete_until,
+           type: :numeric, default: 10_000, desc: "stop deleting files if free space is \
+            above this threshold"
     default_task def watch(root_dir, target_dir)
         root_dir = validate_directory_exists(root_dir)
         target_dir = validate_directory_exists(target_dir)
         archiver = make_archiver(root_dir, target_dir)
         loop do
             archiver.process_root_folder
+            archiver.ensure_free_space(
```
doudou: Call ensure_free_space first to make sure the archiver gets enough free space.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -41,6 +54,36 @@ def process_root_folder
...
+                available_space = stat.bytes_free
+
+                return if available_space > free_space_low_limit
+
+                until available_space >= free_space_delete_until
+                    files = @target_dir.each_child.select(&:file?)
+                    break if files.empty?
+
+                    available_space += files.sort.first.size
+                    files.sort.first.unlink
```
doudou: Do not duplicate long method calls. Give them names (plus, you are actually **sorting** the array twice ... quite wasteful)

```
removed_file = files.sort.first
removed_file.unlink
available_space += removed_file.size
```

doudou: (the inversion of unlink and available_space is to update available_space only after we actually removed the file ... does not change the correctness of the method, just something that I would do personally)

lib/syskit/scripts/log_runtime_archive.rb
```diff
@@ -18,12 +18,21 @@ def self.exit_on_failure?
            type: :numeric, default: 600, desc: "polling period in seconds"
     option :max_size,
            type: :numeric, default: 10_000, desc: "max log size in MB"
+    option :free_space_low_limit,
+           type: :numeric, default: 1_000, desc: "start deleting files if free space is \
+            below this threshold"
+    option :free_space_delete_until,
+           type: :numeric, default: 10_000, desc: "stop deleting files if free space is \
```
doudou: use the constants you have defined in LogRuntimeArchive (actually true for the other parameters)

test/cli/test_log_runtime_archive.rb
```diff
@@ -510,6 +510,99 @@ def should_archive_dataset(dataset, archive_basename, full:)
...
+                end
+
+                it "removes enough files to reach the freed limit" do
+                    size_files = [6, 2, 1, 6, 7, 10, 3, 5, 8, 9]
+                    mock_files_size(size_files)
+                    mock_available_space(0.5)
+
+                    @archiver.ensure_free_space(
+                        free_space_low_limit: @free_space_low_limit,
+                        free_space_delete_until: @free_space_delete_until
```
doudou: These are the defaults. Don't put them explicitly - but write one test where you pass non-default values and check the result.

**or** you just remove the default values passed to `#initialize`. This is probably better actually.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -41,6 +52,33 @@ def process_root_folder
...
+            def ensure_free_space(free_space_low_limit: @free_space_low_limit,
+                free_space_delete_until: @free_space_delete_until)
+                stat = Sys::Filesystem.stat(@target_dir)
+                available_space = stat.bytes_free
+
+                return if available_space > free_space_low_limit
+
+                until available_space >= free_space_delete_until
+                    files = @target_dir.each_child.select(&:file?)
+                    break if files.empty?
```
jhonasiv: maybe warn that the folder is empty but the available space is smaller than the threshold?

test/cli/test_log_runtime_archive.rb
jhonasiv: as Sylvain said, you can just remove the default arguments in this call 
```suggestion
```

lib/syskit/cli/log_runtime_archive.rb
rayssinha-soares: good call... raise an `ArgumentError`?

rayssinha-soares: added Log.warn

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -12,17 +13,27 @@ module CLI
         # It depends on the syskit instance using log rotation
         class LogRuntimeArchive
             DEFAULT_MAX_ARCHIVE_SIZE = 10_000_000_000 # 10G
+            FREE_SPACE_LOW_LIMIT = 5_000_000_000 # 5 G
+            FREE_SPACE_FREED_LIMIT = 25_000_000_000 # 25 G
 
             def initialize(
                 root_dir, target_dir,
                 logger: LogRuntimeArchive.null_logger,
                 max_archive_size: DEFAULT_MAX_ARCHIVE_SIZE
```
doudou: What you did here was actually not what I meant.

Currently, you let the user set instance-wide defaults for free space thresholds **and** override these defaults in the method. I think that's a tad bit too much.

What I meant was that you could remove the instance-wide defaults, keeping only the arguments to ensure_free_space.

(Which makes me realize that the (low_limit < freed_limit check should also be done in ensure_free_space)

lib/syskit/scripts/log_runtime_archive.rb
```diff
@@ -18,11 +18,20 @@ def self.exit_on_failure?
            type: :numeric, default: 600, desc: "polling period in seconds"
     option :max_size,
            type: :numeric, default: 10_000, desc: "max log size in MB"
+    option :FREE_SPACE_LOW_LIMIT,
```
doudou: command line arguments are lowercase. Using the constant was for the 'default' value which is currently not guaranteed to be equal to the class defaults.

lib/syskit/scripts/log_runtime_archive.rb
```diff
@@ -18,11 +18,20 @@ def self.exit_on_failure?
...
+           type: :numeric, default: 5_000, desc: "start deleting files if free space is \
+            below this threshold"
+    option :FREE_SPACE_FREED_LIMIT,
+           type: :numeric, default: 25_000, desc: "stop deleting files if free space is \
+            above this threshold"
```
doudou: Mention the unit in the option description (MB in this case)

lib/syskit/scripts/log_runtime_archive.rb
```diff
@@ -18,11 +18,20 @@ def self.exit_on_failure?
...
     default_task def watch(root_dir, target_dir)
         root_dir = validate_directory_exists(root_dir)
         target_dir = validate_directory_exists(target_dir)
         archiver = make_archiver(root_dir, target_dir)
         loop do
+            archiver.ensure_free_space(
+                options[:FREE_SPACE_LOW_LIMIT], options[:FREE_SPACE_FREED_LIMIT]
```
doudou: Since the limits are MB, you need to multiply by 1000 to get bytes which is what the class expects.

doudou: The method expects keyword arguments.

test/cli/test_log_runtime_archive.rb
```diff
@@ -510,6 +510,87 @@ def should_archive_dataset(dataset, archive_basename, full:)
...
+                    mock_available_space(0.5)
+
+                    @archiver.ensure_free_space(
+                        free_space_low_limit: 1,
+                        free_space_delete_until: 15
+                    )
+                    assert_deleted_files([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+                end
+
+                def mock_files_size(sizes)
```
doudou, line 548: Since `mock_available_space` mocks Sys::Filesystem, you should test the CLI here as well (since it was ... actually buggy).

Move the CLI class from scripts/log_runtime_archive.rb into cli/log_runtime_archive.rb, and then in here you can require the class and call the CLI class as if it was a user from the command line, e.g.

```
CLI::LogRuntimeArchive.start(["log_runtime_archive", "--some", "--arguments", 
     "/as/if/called/from/the/command/line"])
```

The only thing will be to create another temporary directory to pass as the Syskit log dir.

lib/syskit/cli/log_runtime_archive.rb
rayssinha-soares: oh, got it now. fixed.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -41,6 +42,42 @@ def process_root_folder
...
+
+                until available_space >= free_space_delete_until
+                    files = @target_dir.each_child.select(&:file?)
+                    if files.empty?
+                        Roby.warn "Cannot erase files: the folder is empty but the "\
+                        "available space is smaller than the threshold."
+                        break
+                    end
+
+                    removed_file = files.min
```
jhonasiv, line 74: Are we going with removing the smaller file? I thought we were aiming to remove the oldest one, which was also the first one alphabetically given the way we name the logs.

rayssinha-soares: turns out that `.sort.first` (as I was doing before) is equal to `.min` even for strings... actually it was a `Style/RedundantSort` suggestion from Ruby vs code extension and it worked when I tested it

but I can undo this change if you think it's better to understand as it was

jhonasiv: Yeah, it was a misconception from me. String do input the #<=> operator for comparing strings alphabetically.

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -0,0 +1,70 @@
...
+            end
+
+            it "test cli" do
+                size_files = [6, 2, 1, 6, 7, 10, 3, 5, 8, 9]
+                mock_files_size(size_files)
+                mock_available_space(0.5)
+                CLI::LogRuntimeArchiveMain.start(
+                    ["/tmp/archive", "/tmp/datasets",
+                     "--free-space-low-limit", "345526079490",
+                     "--free-space-freed-limit", "345526079500"]
```
jhonasiv: You are using two different paths here (/tmp/.. and /dev/tools/syskit/tmp/dataset). Shouldnt you create a separate tmp path for archive_dir and datasets and then use them?

rayssinha-soares: fixed.

lib/syskit/cli/log_runtime_archive_main.rb
```diff
@@ -0,0 +1,76 @@
...
+            desc "watch", "watch a dataset root folder and call archiver"
+
+            option :period,
+                   type: :numeric, default: 600, desc: "polling period in seconds"
+            option :max_size,
+                   type: :numeric, default: 10_000, desc: "max log size in MB"
+            default_task def watch(root_dir, target_dir)
+                root_dir = validate_directory_exists(root_dir)
+                target_dir = validate_directory_exists(target_dir)
+                loop do
```
doudou: You need to validate in #archive as well. Given the low frequency with which we'll call these methods, doing it only in #archive (removing it from watch) would be good enough.

You should test that the CLI fails if the directory does not exist.

lib/syskit/cli/log_runtime_archive.rb
doudou, line 74: Right ... I am wrong w.r.t. the fact that we can sort the filenames alphabetically.

The file names currently are `<basename>.<number>.tar`, where `number` is not normalized, i.e. `1`, `2`, `3`, `4`... Unfortunately, this means that `somefile.40.log` is lower than `somefile.5.log`, which is rather obviously not what we want.

However, I think the best would be to normalize the archive names to be prefixed with zeroes (e.g. `00040`, `00005`. So, let's keep this PR as-is and do the change in a further PR. We need to have (1) a disk full and (2) more than 100G of logs for that bug to become an issue.... I can live with that.

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -0,0 +1,72 @@
...
+
+                @archiver = LogRuntimeArchive.new(@root, @archive_dir)
+            end
+
+            it "removes enough files to reach the freed limit" do
+                size_files = [6, 2, 1, 6, 7, 10, 3, 5, 8, 9]
+                mock_files_size(size_files)
+                mock_available_space(0.5)
+
+                call_command_line(@root, @archive_dir, 1e-3, 10 * 1e-3)
```
doudou: This is the test's bug.

1e-3 MB is 1 kB. It should have removed all files. You've "cheated" by setting mock_available_space to 0.5, which does not exist (mock_available_space takes bytes as input)

lib/syskit/cli/log_runtime_archive_main.rb
```diff
@@ -0,0 +1,76 @@
...
+                    available space is below this threshold (threshold in MB)"
+            option :free_space_freed_limit,
+                   type: :numeric, default: 25_000, desc: "stop deleting files if \
+                    available space is above this threshold (threshold in MB)"
+            def archive(root_dir, target_dir)
+                archiver = make_archiver(root_dir, target_dir)
+
+                archiver.ensure_free_space(
+                    options[:free_space_low_limit] * 1000,
+                    options[:free_space_freed_limit] * 1000
```
doudou: Right. The units do not match. `ensure_free_space` expects bytes, `free_space_low_limit` is documented to be in MB. The error is here *and* in the tests (there's a comment in the test)

test/cli/test_log_runtime_archive_main.rb
rayssinha-soares: fixed

lib/syskit/cli/log_runtime_archive_main.rb
rayssinha-soares: fixed.

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -0,0 +1,105 @@
...
+                    assert_deleted_files([])
+                end
+
+                it "removes enough files to reach the freed limit" do
+                    size_files = [750, 400, 900, 600, 700]
+                    mock_files_size(size_files)
+                    mock_available_space(700) # 700 MB
+
+                    call_command_line(@root, @archive_dir, 1000, 3000)
+                    assert_deleted_files([0, 1, 2, 3])
```
jhonasiv: You increased the size too much IMO, maybe just remove a 0 from everyone here

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -0,0 +1,105 @@
...
+                end
+
+                it "stops removing files when there is no file in folder even if freed
+                    limit is not achieved" do
+                    size_files = Array.new(5, 100)
+                    mock_files_size(size_files)
+                    mock_available_space(500)
```
jhonasiv: Increase the available space to something like 800 MB? That ensures it actually tests what it should.

doudou (approved):

rayssinha-soares:
This PR passed in CI (https://buildbot.tidewise.io/#/builders/4893/builds/32) and it's able to merge @doudou @jhonasiv

jhonasiv:
> This PR passed in CI (https://buildbot.tidewise.io/#/builders/4893/builds/32) and it's able to merge @doudou @jhonasiv

Merging it then

PR #389, opened 2024-03-03
fix: log_runtime_archive `watch` subcommand

doudou:
The name in scripts/log_runtime_archive_main did not match the definition in cli/log_runtime_archive_main. Pick the name that maches the file name.

### DIFF
--- a/lib/syskit/cli/log_runtime_archive_main.rb
+++ b/lib/syskit/cli/log_runtime_archive_main.rb
@@ -29,7 +29,11 @@ module Syskit
                     available space is above this threshold (threshold in MB)"
             default_task def watch(root_dir, target_dir)
                 loop do
-                    archive(root_dir, target_dir)
+                    begin
+                        archive(root_dir, target_dir)
+                    rescue Errno::ENOSPC
+                        next
+                    end
 
                     puts "Archived pending logs, sleeping #{options[:period]}s"
                     sleep options[:period]
--- a/lib/syskit/cli/log_runtime_archive_main.rb
+++ b/lib/syskit/cli/log_runtime_archive_main.rb
@@ -21,6 +21,12 @@ module Syskit
                    type: :numeric, default: 600, desc: "polling period in seconds"
             option :max_size,
                    type: :numeric, default: 10_000, desc: "max log size in MB"
+            option :free_space_low_limit,
+                   type: :numeric, default: 5_000, desc: "start deleting files if \
+                    available space is below this threshold (threshold in MB)"
+            option :free_space_freed_limit,
+                   type: :numeric, default: 25_000, desc: "stop deleting files if \
+                    available space is above this threshold (threshold in MB)"
             default_task def watch(root_dir, target_dir)
                 loop do
                     archive(root_dir, target_dir)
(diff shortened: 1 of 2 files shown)


doudou:
@rayssinha-soares FYI

lib/syskit/cli/log_runtime_archive_main.rb
rayssinha-soares, line 1: with the options description in `watch` block, can you omit from the `archive` block?

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -112,6 +101,44 @@ def call_archive(root_path, archive_path, low_limit, freed_limit)
                     )
                 end
             end
+
```
rayssinha-soares, line 129: Is there a reason/coding standard for change this order?

lib/syskit/cli/log_runtime_archive_main.rb
rayssinha-soares, line 1: with the options description in `watch` block, can you omit from the `archive` block?

test/cli/test_log_runtime_archive_main.rb
rayssinha-soares, line 130: Is there a reason/coding standard for this order change?

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -6,30 +6,94 @@
...
+                    end
+                    assert_equal "#{root} does not exist, or is not a directory", e.message
+                end
 
-                    @archiver = LogRuntimeArchive.new(@root, @archive_dir)
+                it "raises ArgumentError if the target directory does not exist" do
+                    e = assert_raises ArgumentError do
+                        call_archive(@root, "/does/not/exist", 1, 10)
+                    end
+                    assert_equal "#{root} does not exist, or is not a directory", e.message
```
rayssinha-soares:
```suggestion
                it "raises ArgumentError if the target directory does not exist" do
                    target = "/does/not/exist"
                    e = assert_raises ArgumentError do
                        call_archive(@root, target, 1, 10)
                    end
                    assert_equal "#{target} does not exist, or is not a directory", e.message
```

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -6,30 +6,94 @@
...
                     5.times { |i| (@archive_dir / i.to_s).write(i.to_s) }
+                end
+
+                it "raises ArgumentError if the source directory does not exist" do
+                    root = "/does/not/exist"
+
+                    e = assert_raises ArgumentError do
+                        call_archive(@root, "/does/not/exist", 1, 10)
+                    end
+                    assert_equal "#{root} does not exist, or is not a directory", e.message
```
rayssinha-soares:
```suggestion
                it "raises ArgumentError if the source directory does not exist" do
                    root = "/does/not/exist"

                    e = assert_raises ArgumentError do
                        call_archive(root, @archive_dir, 1, 10)
                    end
                    assert_equal "#{root} does not exist, or is not a directory", e.message
```

test/cli/test_log_runtime_archive_main.rb
doudou: No. This is a change of scope. I moved the common helpers into the base `describe` block and now have `call_archive` in the archive `describe` block.

lib/syskit/cli/log_runtime_archive_main.rb
doudou: Nope. `archive` and `watch` are separate commands. You need them to be defined in `archive` so that the `archive` command works.

We also need to set them in `watch` because `watch` calls `archive`

doudou:
Ping @rayssinha-soares

rayssinha-soares (approved):

PR #390, opened 2024-03-13
fix: use bytes_available instead of bytes_free to determine free space

doudou:
The former is the total amount free on disk, the latter takes into account space reserved for priviledged users (e.g. root)


doudou:
FYI @rayssinha-soares

jhonasiv (approved):

rayssinha-soares (approved):

PR #391, opened 2024-03-13
fix: do not "eat" exceptions in the log archiving process

doudou:
The current scheme was trying very hard to compress other files, assuming that the error would be file-specific. Turns out that this case does not happen (at least, did not happen so far) but it's very common to have a fatal error (think: no space on disk).

### DIFF
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -5,6 +5,9 @@ require "sys/filesystem"
 
 module Syskit
     module CLI
+        # Exception raised when the zstd subprocess returns an error
+        class CompressionFailed < RuntimeError; end
+
         # Implementation of the `syskit log-runtime-archive` tool
         #
         # The tool archives Syskit log directories into tar archives in realtime,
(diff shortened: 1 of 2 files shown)


doudou:
FYI @rayssinha-soares

jhonasiv (approved):

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -170,22 +173,20 @@ def self.add_to_archive(archive_io, child_path, logger: null_logger)
...
+                    archive_io, child_path, start_pos, data_pos, stat
+                )
+                child_path.unlink
             rescue Exception => e # rubocop:disable Lint/RescueException
                 Roby.display_exception(STDOUT, e)
                 if start_pos
                     add_to_archive_rollback(archive_io, start_pos, logger: logger)
                 end
-                false
+                raise
```
rayssinha-soares, line 189: I didn't get which error this `raise` will show... is it the same from [line 177](https://github.com/rock-core/tools-syskit/pull/391/files#diff-27a24ddb639c1c67b459feffe67d0e7556ab8a2dba9de22165c32aef63d9fb4bR177)?

doudou: raise without arguments raises the last rescued exception (in this case, `e`)

PR #394, opened 2024-03-18
fix: jobs may not be action tasks

doudou:


wvmcastro (commented):
1. Can you remind me when a job does not come from an action? 
2. So, when a job doesn't come from an action, we will be able to drop it but not start/restart?

doudou:
> 1. Can you remind me when a job does not come from an action?

Whenever. A job is only a task that provides Actions::Job and has a non-nil job_id

> 2. So, when a job doesn't come from an action, we will be able to drop it but not start/restart?

Yes, since we don't have a standard interface to start or restart it

wvmcastro (approved):

PR #395, opened 2024-03-18
fix: handle resolution raising an error while all InstanceRequirements task have been removed

doudou:
Depends on:
- [ ] https://github.com/rock-drivers/drivers-orogen-controldev_websocket/pull/13
- [ ] https://github.com/rock-drivers/drivers-orogen-controldev_websocket/pull/15

This was triggering a (wrong) sanity check in the async resolution error handling, that was in the end killing syskit.

### DIFF
--- a/test/runtime/test_apply_requirement_modifications.rb
+++ b/test/runtime/test_apply_requirement_modifications.rb
@@ -44,7 +44,33 @@ module Syskit
                 execute { Runtime.apply_requirement_modifications(plan) }
 
                 assert_resolution_cancelled do
-                    execute { plan.unmark_permanent_task(requirement_task) }
+                    expect_execution do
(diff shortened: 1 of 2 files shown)


test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -44,7 +44,35 @@ module Runtime
                 execute { Runtime.apply_requirement_modifications(plan) }
 
                 assert_resolution_cancelled do
-                    execute { plan.unmark_permanent_task(requirement_task) }
+                    expect_execution do
```
wvmcastro, line 47: Can you explain what `assert_resolution_cancelled` does? I didn't quite get reading the code

wvmcastro (approved):

PR #398, opened 2024-04-05
feat: interface v2 protocol

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/275

The integration is done as to not break the v1 protocol. The v2 protocol marshallers are registered only if using the v2 interface, using the app's setup_interface_v2_protocol hook

The biggest change is a plain copy (without changes) of syskit/shell_commands to syskit/interface/commands. The rest is minimal.


doudou:
Ping

lib/syskit/interface/commands.rb
wvmcastro, line 1: Is it really without changes? _roby_app/shell_inerface_ and _interface/commands_ have different number of lines. Also, I think `git mv` would really do the job here decreasing the diff size

wvmcastro (approved):

lib/syskit/interface/commands.rb
jhonasiv: It seems he added some deployment-related functionalities, like the deployment command and moved some code to functions for restart_deployments

jhonasiv (approved):

lib/syskit/interface/commands.rb
doudou: `git mv` is strictly equal to `git rm` and `git add`. Git does not have the concept of "moves" per se. It detects copies and moves by textual comparison.

In this case, I believe the issue is with Github, which does not represent copies and moves.

PR #400, opened 2024-04-05
feat: implement `syskit telemetry ui` based on the copy of the current runtime view

doudou:
On top of #399

PR #401, opened 2024-04-05
feat: use the v2 interface in `telemetry ui`

doudou:
On top of #400 

This uses Syskit to resolve orocos tasks, and uses the new protocol to show data. At least, it tremendously speeds up task resolution, and ensures that we see all the tasks Syskit sees (instead of only the tasks that are present on the main computer)

It removes the use / display of the task log, which does remove some information. This will be fixed in future PRs.

[sc-43835] [sc-43836] [sc-43837]

### DIFF
--- a/lib/syskit/interface/v2/protocol.rb
+++ b/lib/syskit/interface/v2/protocol.rb
@@ -1,21 +1,36 @@
 # frozen_string_literal: true
 
+require "roby/interface/v2"
+
 module Syskit
     module Interface
         module V2
             # Syskit extensions to Roby's v2 interface wire protocol
             module Protocol
+                ROBY_TASK_MEMBERS = Roby::Interface::V2::Protocol::Task.new.members
+
                 Deployment = Struct.new(
-                    :roby_task, :pid, :ready_since, :iors, keyword_init: true
+                    *ROBY_TASK_MEMBERS,
+                    :pid, :ready_since, :deployed_tasks, keyword_init: true
                 ) do
                     def pretty_print(pp)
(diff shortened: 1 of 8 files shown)


lib/syskit/telemetry/ui/name_service.rb
```diff
@@ -0,0 +1,68 @@
...
+                    @registered_tasks = Concurrent::Hash.new
+                    tasks.each { |task| register(task) }
+                end
+
+                def names
+                    @registered_tasks.keys
+                end
+
+                def include?(name)
+                    puts @registered_tasks.keys
```
jhonasiv: This `puts` doesnt seem intentional

lib/syskit/interface/v2/protocol.rb
jhonasiv, line 1: Is this supposed to be in this PR? I seems like it should be done in another. I dont mind it staying since you are dealing with a bunch of "On top of" features, but I feel it needs more context, maybe a separate commit?

doudou: This is a protocol modification that was done for the benefit of the runtime-view. I'll move it to a separate commit. Don't see the value of separating in another PR.

doudou: Splitting done.

lib/syskit/telemetry/ui/name_service.rb
doudou: It was already removed on my branch ... not sure what happened here.

jhonasiv (approved):

PR #404, opened 2024-04-09
feat: inject class on data readers

jhonasiv:


lib/syskit/models/component.rb
```diff
@@ -1311,14 +1311,18 @@ def match
             #    data_reader some_child.out_port, as: 'pose'
             #
             # @return [DynamicPortBinding::BoundOutputReader]
-            def data_reader(port, as:, **policy)
+            def data_reader(
```
wvmcastro, line 1325: Add example and doc (maybe?)

wvmcastro (approved):

lib/syskit/models/dynamic_port_binding.rb
```diff
@@ -197,9 +197,24 @@ def initialize(name, component_model, *arguments, **kw_arguments)
...
+                    klass: Syskit::DynamicPortBinding::BoundOutputReader, **kw_arguments
+                )
+                    super(name, component_model, *arguments, **kw_arguments)
+
+                    unless klass <= Syskit::DynamicPortBinding::BoundOutputReader
+                        raise ArgumentError,
+                              "#{klass} does not subclass " \
+                              "Syskit::DynamicPortBinding::BoundOutputReader"
+                    end
+
```
doudou: This is Ruby. That's not your problem (specifically at this level of the usage of the API)

test/models/test_dynamic_port_binding.rb
```diff
@@ -455,6 +455,42 @@ module Models
...
+                            "bla", flexmock, @port, klass: foo
+                        )
+                    end
+                end
+
+                it "instantiates using BoundOutputReader when klass is not passed" do
+                    b = DynamicPortBinding::BoundOutputReader.new("bla", flexmock, @port)
+                    result = b.instanciate(flexmock)
+                    assert(
+                        result.instance_of?(Syskit::DynamicPortBinding::BoundOutputReader)
```
doudou: I think rubocop wants `kind_of?` (generally speaking, I use kind_of? so please stick with that)

test/models/test_dynamic_port_binding.rb
```diff
@@ -455,6 +455,42 @@ module Models
...
+                it "instantiates using the injected class" do
+                    klass = Class.new Syskit::DynamicPortBinding::BoundOutputReader
+
+                    b = DynamicPortBinding::BoundOutputReader.new(
+                        "bla", flexmock, @port, klass: klass
+                    )
+                    result = b.instanciate(flexmock)
+                    assert(
+                        result.kind_of?(Syskit::DynamicPortBinding::BoundOutputReader)
+                    )
```
doudou: This makes little sense ... if result.kind_of?(klass) and `klass <= Syskit::DynamicPortBinding::BoundOutputReader` then it is inferred.

test/models/test_dynamic_port_binding.rb
doudou: Plus, I believe minitest has `assert_kind_of`

doudou (approved):

PR #407, opened 2024-05-20
fix: allow passing a Syskit model as a deployment task's model (on top of #409)

doudou:
On top of #409 

We were currently passing the orogen model, which was resolved
to a Syskit model using TaskContext.model_for

This is based on the obvious assumption that only one Syskit model
may ever have a given orogen model, which is broken by syskit-log.
In addition, it would not protect the syskit models against Ruby's
garbage collector, which is I believe why we were having random
failures in tests for Syskit models that did not exist.

### DIFF
--- a/lib/syskit/roby_app/configuration.rb
+++ b/lib/syskit/roby_app/configuration.rb
@@ -81,6 +81,13 @@ module Syskit
             # @deprecated Unused, kept here for historical reasons
             attr_predicate :use_only_model_pack?, true
 
+            # Controls whether Models::TaskContext.model_for raises if there is more
+            # than one Syskit model matching the given orogen model
+            #
+            # This is false for backward compatibility reasons, and will eventually
+            # become the default behavior.
+            attr_predicate :strict_model_for, true
+
             # Data logging configuration
             #
             # @return [LoggingConfiguration]
@@ -144,6 +151,7 @@ module Syskit
                 @exception_transition_timeout = 20.0
                 @kill_all_on_process_server_connection = false
                 @register_self_on_name_server = (ENV["SYSKIT_REGISTER_SELF_ON_NAME_SERVER"] != "0")
+                @strict_model_for = false
 
                 @log_rotation_period = nil
                 @log_transfer = LogTransferManager::Configuration.new(
--- a/test/models/test_deployment.rb
+++ b/test/models/test_deployment.rb
@@ -74,20 +74,38 @@ module Syskit
     end
 end
 
-describe Syskit::Models::Deployment do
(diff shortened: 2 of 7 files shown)


wvmcastro (approved):

doudou:
Ping @wvmcastro

lib/syskit/roby_app/configuration.rb
```diff
@@ -81,6 +81,13 @@ class Configuration
             # @deprecated Unused, kept here for historical reasons
             attr_predicate :use_only_model_pack?, true
 
+            # Controls whether Models::TaskContext.model_for raises if there is more
+            # than one Syskit model matching the given orogen model
+            #
+            # This is false for backward compatibility reasons, and will eventually
+            # become the default behavior.
+            attr_predicate :strict_model_for, true
```
jhonasiv, line 89: Why not set the default to false here instead of doing so in `#initialize`? This is very minor, but it could lead to some confusion when reading the documentation

test/models/test_deployment.rb
```diff
@@ -29,28 +29,128 @@ def test_new_submodel_registers_the_submodel
...
+                    task "task2", task2_m.orogen_model
+                end
+
+                assert_equal [task1_m.orogen_model, task2_m.orogen_model],
+                             deployment_m.each_orogen_deployed_task_context_model
+                                         .sort_by(&:name).map(&:task_model)
+                assert_equal [["task1", task1_m], ["task2", task2_m]],
+                             deployment_m.each_deployed_task_model
+                                         .sort_by(&:first)
+            end
```
jhonasiv, line 48: duplicate test

test/models/test_deployment.rb
```diff
@@ -29,28 +29,128 @@ def test_new_submodel_registers_the_submodel
...
+
+            it "lets the caller define the deployment model, "\
+               "using orogen model names for the tasks" do
```
jhonasiv, line 51: same name as the previous tests

lib/syskit/roby_app/configuration.rb
doudou: The boolean argument to `attr_predicate` says whether the predicate is readonly (false) or readwrite (true). It`s not a default value.

test/models/test_deployment.rb
doudou: removed

lib/syskit/roby_app/configuration.rb
jhonasiv: Ah I see, did not know about that

jhonasiv (approved):

PR #409, opened 2024-05-30
fix: wait for the channel to be ready when establishing a data channel

doudou:


test/telemetry/agent/test_client_server.rb
```diff
@@ -182,20 +182,21 @@ def wait_locked(timeout)
...
                             enum.each do |sample|
-                                sample_queue.enq(sample)
+                                # We ignore the first sample. It is sent as a way to
+                                # confirm that the channel is successfully established
+                                sample_queue.enq(sample) unless channel_ready.set?
                             end
                         rescue GRPC::Core::CallError, GRPC::Cancelled # rubocop:disable Lint/SuppressedException
                         end
-                        channel_ready.wait
+                        channel_ready.wait(timeout)
```
jhonasiv: Where is channel_ready being set? Shouldnt it be done by the first sample reception?

doudou: Good catch.

The set? above was meant to be a try? (which sets only if not yet set, and returns true in that case). And `Event#wait(timeout)` returns a boolean. I was assuming it to raise...

Both fixed.

jhonasiv (approved):

PR #412, opened 2024-06-05
chore: remove unused task variable

eduardacoppo:
Rubocop detects this as an offense


doudou:
This is on purpose. The goal is for people to at least open the file and decide what to do.

PR #414, opened 2024-07-23
feat: implement deployment of orogen-generated components in-process (on top of #413)

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-orogen_syskit_tests/pull/6
- [ ] https://github.com/rock-core/tools-orocosrb/pull/156
- [ ] https://github.com/rock-core/tools-orogen/pull/17
- [ ] https://github.com/rock-core/tools-syskit-log/pull/48

On top of #413

By "in-process", one means that the C++ component runs inside the Syskit process.

The functionality is accessed by declaring a deployment using `using_in_process_task TaskModel => "name"`


lib/syskit/roby_app/in_process_tasks_manager.rb
```diff
@@ -0,0 +1,163 @@
...
+    module RobyApp
+        # A class API-compatible with {Syskit::RobyApp::RemoteProcesses::Client} but that
+        # creates components in-process
+        class InProcessTasksManager
+            # Exception raised if one attempts to do name mappings in an
+            # unmanaged process server
+            class NameMappingsForbidden < ArgumentError; end
+
+            # Fake status class
+            class Status
```
jhonasiv, line 13: I dont see the point of this class, on wait_termination you mention returning a mapping with the status but in fact you are returning a mapping with the process. Also, isnt it possible to get the actual status with the in process task if its terminated?

doudou: That's ... a good catch.

Essentially, wait_termination does not return any status. This actually causes every deployment using this process manager to emit failed on stop. The unmanaged task manager has the same issue.

doudou: Fixed

wvmcastro (approved):

jhonasiv (approved):

PR #416, opened 2024-07-24
feat: implement logging ruby tasks (on top of #414)

doudou:
On top of #414 

This PR implements using an in-process logger for ruby tasks. The logger is disabled by default (as it would have the ability to seriously impact a system in case of bugs). One needs to set

    Syskit.conf.in_process_default_logger_support = true

to enable it.


jhonasiv (commented):
Just so that I understand, how would having a single deployment for all the in process tasks work when we have multiple in process tasks running?

doudou:
> Just so that I understand, how would having a single deployment for all the in process tasks work when we have multiple in process tasks running?

Not sure I understand the question.

The mapping between deployments and in-process tasks is "free". That is, one can have more than one in one "deployment" or a single one. Obviously, just like with the ruby tasks, the deployment is virtual (it's only there to manage whether or not the task objects are created)

jhonasiv:
> > Just so that I understand, how would having a single deployment for all the in process tasks work when we have multiple in process tasks running?
> 
> Not sure I understand the question.
> 
> The mapping between deployments and in-process tasks is "free". That is, one can have more than one in one "deployment" or a single one. Obviously, just like with the ruby tasks, the deployment is virtual (it's only there to manage whether or not the task objects are created)

That answers my question, thanks.

jhonasiv (approved):

PR #428, opened 2024-08-12
integrate `rr` as an execution mode (on top of #427)

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-orogen/pull/20

On top of #427 

rr is a reverse debugger (https://rr-project.org/). It allows to record the state of a program while it is being executed, and then debug back-and-forth deterministically over and over again.

This PR implements `rr` as an execution mode. Simply do

`deployed_as("name", execution_mode: { type: "rr" })`

The recorded states are saved in the log folder.

### DIFF
--- a/lib/syskit/process_managers/remote/server/process.rb
+++ b/lib/syskit/process_managers/remote/server/process.rb
@@ -370,7 +370,7 @@ module Syskit
                         JSON.parse(message)
                     rescue JSON::ParserError
                         raise Orocos::InvalidIORMessage,
-                              "received IOR message is not valid JSON"
+                              "received IOR message is not valid JSON: #{message}"
                     end
 
                     SIGNAL_NUMBERS = {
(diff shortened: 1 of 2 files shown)


lib/syskit/process_managers/remote/server/process.rb
```diff
@@ -250,14 +266,28 @@ def spawn
                         @ior_message = +""
 
                         @ior_read_fd, ior_write_fd = IO.pipe
+                        control_read_fd, @control_write_fd = IO.pipe
```
jhonasiv, line 269: Why does the `rr` execution mode implementation need to use the control fd instead of signal handling?

doudou: It is discussed in the control fd implementation

Essentially, support for signals in rr is sketchy. But, to be honest, signals are really not that great in multithreaded code.

jhonasiv: > signals are really not that great in multithreaded code.

I figured, just wondered if there was any special requirements for rr.

jhonasiv (approved):

PR #433, opened 2024-09-03
fix: make sure LogUploadState is defined in Remote::Manager

doudou:
It is sent by the server, and if not defined in the Syskit process, leads to a failure at runtime (type does not exist).

### DIFF
--- a/lib/syskit.rb
+++ b/lib/syskit.rb
@@ -42,6 +42,7 @@ end
 require "syskit/roby_app/log_transfer_server"
 require "syskit/process_managers/process_base"
 require "syskit/process_managers/status"
+require "syskit/process_managers/remote/server/log_upload_state"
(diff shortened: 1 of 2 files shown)


wvmcastro (approved):
I wonder why it does not break any test

doudou:
Because of this

```
            # Defined here to make sure it is actually defined. Otherwise, the log
            # state reporting would fail at runtime, and unit-testing for this is
            # very hard.
```

The type is defined during tests because the server code is required within the process that runs the syskit test suite (since that process does run the server code).

The only way to ensure the test suite would catch it is to run a syskit instance separately with the log transfer enabled and check for log transfer ... I preferred adding a reference to the type in the Manager class to make sure it does not happen again.

The "real fix" would be to start using a proper RPC implementation instead of this shitty handmade protocol ... but that's for another PR ;-)

PR #441, opened 2024-12-04
[WIP] feat: copy log to central system

eduardacoppo:


lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -45,6 +49,32 @@ def process_root_folder
...
+                    mng.transfer(child)
+                end
+            end
+
+            def log_transfer_prepare
+                return unless Syskit.conf.log_transfer.ip
+
+                conf = Syskit.conf.log_transfer
+                conf.target_dir ||= log_dir
+                @log_transfer_manager = LogTransferManager.new(conf)
```
doudou: You won't have access to these configurations within log_runtime_archive. It does not load a robot configuration. Need to pass all arguments as command line.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -45,6 +49,32 @@ def process_root_folder
...
+            # @param [Pathname] src_dir the log folder on the process server
+            # @param [Pathname] dst_dir the folder on the main computer
+            #   in which to transfer the logs to
+            def process_transfer(src_dir, dst_dir)
+                return unless (mng = log_transfer_manager)
+
+                candidates = self.class.find_all_dataset_folders(@src_dir)
+                candidates.each do |child|
+                    mng.transfer(child)
+                end
```
doudou: The general structure is fine, but this is very wrong.

From a coding perspective, you are passing src_dir as argument *and* as instance variable ? You should be picking these coding errors just by looking at them.

From an architectural perspective, the log transfer manager is not usable as-is anymore in this work. In the original syskit structure, we were delegating the log transfer to the process servers (which is what `#transfer` is doing, calling the process servers to do the transfers). In this structure, we won't be doing this anymore. Instead the log runtime archive process *on the remote machines* should initiate transfers when needed.

To decide what to transfer, you need to refactor the logic in log_runtime_archive to separate the code that decides what to archive with the code that actually does the archiving, so that you can reuse the first part to transfer instead of archive. The link is in shortcut.

lib/syskit/roby_app/plugin.rb
```diff
@@ -33,10 +33,6 @@ module RobyApp
         #   loaded models. In addition, a text notification is sent to inform
         #   a shell user
         module Plugin
-            # @return [LogTransferManager,nil] the log transfer support object, or nil
-            #   before {.setup}, or if log transfer is not enabled
-            attr_accessor :syskit_log_transfer_manager
-
```
doudou: Do not remove the old code. We'll remove it when the new mechanism is validated.

The old implementation is better than nothing.

lib/syskit/cli/log_runtime_archive.rb
eduardacoppo: > you are passing src_dir as argument and as instance variable ?

I was indeed using it as an instance variable before passing it as argument. Must have overlooked this while cleaning the code.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -147,6 +160,23 @@ def find_last_archive_index(basename)
...
 
+            def self.create_tcp_socket(host, port)
+                socket =
+                    begin TCPSocket.new(host, port)
+                    rescue Errno::ECONNREFUSED => e
+                        raise e.class, "cannot contact process server at " \
+                                       "'#{host}:#{port}': #{e.message}"
+                    end
+                socket.write(ProcessManagers::Remote::COMMAND_LOG_UPLOAD_FILE)
+                socket
```
jhonasiv: @doudou correct me if I am wrong, but she will need to use a Net::FTP client, right? I had told her to base herself on what the process server is doing, but this seems wrong now that I looking at it.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -81,26 +99,21 @@ def ensure_free_space(free_space_low_limit, free_space_delete_until)
...
-                            io, child,
-                            logger: @logger, full: full,
-                            max_size: @max_archive_size
-                        )
-                        return if dataset_complete
-                    end
+            def transfering_and_archiving_candidates(candidates)
+                transfer, archive = candidates.partition do |child|
+                    self.class.archive_size_exceeded?(child)
+                end
```
jhonasiv: Why do you need to partition? Is there a different policy between transfer and archiving that I am not getting?

lib/syskit/cli/log_runtime_archive_main.rb
```diff
@@ -61,6 +61,42 @@ def archive(root_dir, target_dir)
                 archiver.process_root_folder
             end
 
+            desc "watch_transfer", "watches a dataset root folder \
+                                    and periodically performs transfer"
+            option :period,
+                   type: :numeric, default: 600, desc: "polling period in seconds"
+            option :max_size,
+                   type: :numeric, default: 10_000, desc: "max log size in MB"
+            default_task def watch_transfer(src_dir, tgt_dir, server_params)
```
wvmcastro: The `tgt_dir` parameter makes no sense for `transfer` and `watch_transfer` commands because they don't do the archiving

Also, here in `watch_transfer` I don't like to just call `transfer` because it creates an archiver and does socket allocation every time

lib/syskit/cli/log_runtime_archive.rb
eduardacoppo: @jhonasiv what exactly seems off here?

wvmcastro (commented):
I am not sure you got the change between the current log transfer and archiving schema and the new one we are moving towards too. Just to be sure we are all at the same page

Current: There is a central node (LogTransferManager) that runs in the master machine and periodically tells the slave machines _process servers_ to send their logs to a target dir. Then, another tool (LogRuntimeArchive) gets the data accumulated in this target folder and does the archiving (compress and move to another storage)

New: The LogTransferManager is taken out the dynamic, and now the slave machines will have a client that periodically sends its logs to a LogTransferServer at the master machine. Then, the master machine LogRuntimeArchive continues to do what it currently does

Given that, I think a great way to test your implementation is to setup a server in your machine and a use a client to get files from on folder to another

eduardacoppo:
> I am not sure you got the change between the current log transfer and archiving schema and the new one we are moving towards too. Just to be sure we are all at the same page
> 
> Current: There is a central node (LogTransferManager) that runs in the master machine and periodically tells the slave machines _process servers_ to send their logs to a target dir. Then, another tool (LogRuntimeArchive) gets the data accumulated in this target folder and does the archiving (compress and move to another storage)
> 
> New: The LogTransferManager is taken out the dynamic, and now the slave machines will have a client that periodically sends its logs to a LogTransferServer at the master machine. Then, the master machine LogRuntimeArchive continues to do what it currently does
> 
> Given that, I think a great way to test your implementation is to setup a server in your machine and a use a client to get files from on folder to another

This is clear to me. The biggest difficulty lies in implementing it.

lib/syskit/cli/log_runtime_archive_main.rb
eduardacoppo: > Also, here in watch_transfer I don't like to just call transfer because it creates an archiver and does socket allocation every time

What do you suggest? I decided to keep the names suggested in the shortcut story.

jhonasiv: She does need a target directory to transfer the files to.

jhonasiv: Also, I dont like the CLI interface for this command and the transfer. Server params shouldnt be a positional argument, it is a hash, it needs to be split into separate options  (host, port and anything else that would be included)

lib/syskit/cli/log_runtime_archive.rb
jhonasiv: I answered her in Chat, but for information's sake. This was used by process server to talk to each other, which is obviously not the case anymore.

lib/syskit/cli/log_runtime_archive_main.rb
eduardacoppo: > The tgt_dir parameter makes no sense for transfer and watch_transfer commands because they don't do the archiving

Right. I do plan on changing the initialize function so I can have the option to not pass a `target_dir` if not needed.

eduardacoppo: > Also, I dont like the CLI interface for this command and the transfer. Server params shouldnt be a positional argument, it is a hash, it needs to be split into separate options (host, port and anything else that would be included)

Could you elaborate? I'm not sure if that's what you meant: having a variable for each value (host, port, etc.) and passing them as arguments to the `process_transfer` function? How should I deal with rubocop then (he doesn't like functions with more than 5 arguments), if that's indeed what you meant?

jhonasiv: I mean on the interface one will use to interact with these commands (aka each one of the "default_task"s you created). You cant pass a hash via command line, my suggestion is that the server_params become individual options.

doudou: And you really should not write *anything* without understanding. This smells like blind copy/pasting.

What does `default_task` does ?

doudou:
> I am not sure you got the change between the current log transfer and archiving schema and the new one we are moving towards too. Just to be sure we are all at the same page
> 
> Current: There is a central node (LogTransferManager) that runs in the master machine and periodically tells the slave machines _process servers_ to send their logs to a target dir. Then, another tool (LogRuntimeArchive) gets the data accumulated in this target folder and does the archiving (compress and move to another storage)
> 
> New: The LogTransferManager is taken out the dynamic, and now the slave machines will have a client that periodically sends its logs to a LogTransferServer at the master machine. Then, the master machine LogRuntimeArchive continues to do what it currently does
> 
> Given that, I think a great way to test your implementation is to setup a server in your machine and a use a client to get files from on folder to another

There are unit tests for the current log transfer functionality. There should be unit tests for the new functionality as well.

lib/syskit/cli/log_runtime_archive_main.rb
eduardacoppo: > I mean on the interface one will use to interact with these commands (aka each one of the "default_task"s you created). You cant pass a hash via command line, my suggestion is that the server_params become individual options.

Ok. I'll still have the rubocop problem, but I guess there's no way to overcome this. Will disable the cop, then.

lib/syskit/cli/log_runtime_archive_main.rb
```diff
@@ -61,6 +61,42 @@ def archive(root_dir, target_dir)
...
+
+                    puts "Transferred pending logs, sleeping #{options[:period]}s"
+                    sleep options[:period]
+                end
+            end
+
+            desc "transfer", "transfers the datasets"
+            option :max_size,
+                   type: :numeric, default: 10_000, desc: "max log size in MB"
+            def transfer(src_dir, tgt_dir, server_params)
```
jhonasiv: src_dir -> base_log_dir

jhonasiv: tgt_dir will become the last folder inside the path for each dataset. It will be available inside process_transfer, when you are iterating on the dataset candidates, each candidate path (which is the timestamp in our case) is the tgt dir

eduardacoppo:
Work continued in this PR: https://github.com/rock-core/tools-syskit/pull/448

PR #442, opened 2024-12-04
chore: propage permanent mark to task instance generated from deployment

wvmcastro:
It is not strictly needed since the deployment task instance is added to the plan after generation. But as we pretend to do early deployment at generation, the garbage collection would throw away tasks contexts that got merged into this instances.


wvmcastro:
Not really the right way to go, replace for https://github.com/rock-core/tools-syskit/pull/443

PR #443, opened 2024-12-05
 chore: copy plan marks when applying merge group

wvmcastro:
Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/304

This became a necessity as we pretend to do early deployments at network generation. In this scenario, when a "free" task context is merged into its deployed counter part its plan marks is not carried out to the deployment task instance. Then, this task instance would be garbage collected because it did not carried the permanent mark from the replace "free" task context

https://buildbot.tidewise.io/#/builders/1791/builds/15

### DIFF
--- a/lib/syskit/network_generation/merge_solver.rb
+++ b/lib/syskit/network_generation/merge_solver.rb
@@ -138,6 +138,7 @@ module Syskit
 
                 merged_task_to_task.each do |merged_task, task|
                     unless merged_task.transaction_proxy?
+                        plan.copy_task_marks(from: merged_task, to: task)
                         plan.remove_task(merged_task)
                     end
                     register_replacement(merged_task, task)
--- a/test/network_generation/test_merge_solver.rb
+++ b/test/network_generation/test_merge_solver.rb
@@ -358,6 +358,31 @@ describe Syskit::NetworkGeneration::MergeSolver do
         end
     end
 
+    describe "#apply_merge_group" do
+        attr_reader :local_plan, :solver
+
+        before do
+            @local_plan = Roby::Plan.new
+            @solver = Syskit::NetworkGeneration::MergeSolver.new(@local_plan)
+        end
+
+        it "applyes merged task plan marks to the destination task" do
+            task1 = Roby::Task.new
+            task2 = Roby::Task.new
+
+            local_plan.add_permanent_task task1
+            local_plan.add_mission_task task1
+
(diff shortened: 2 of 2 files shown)


lib/syskit/network_generation/merge_solver.rb
```diff
@@ -138,6 +138,7 @@ def apply_merge_group(merged_task_to_task)
 
                 merged_task_to_task.each do |merged_task, task|
                     unless merged_task.transaction_proxy?
+                        plan.copy_marks! from_task: merged_task, to_task: task
```
doudou: I think the copy should happen even if merged_task is a transaction proxy

wvmcastro: I can't find what a _transaction proxy_ really is. Can you point me somewhere I can look for?

doudou: ... my PhD probably.

Essentially, a transaction is a plan-like structure that represents a change in the plan. It is either committed (all changes are applied) or discarded (no changes are applied). This is what the work_plan actually is.

A transaction proxy is the representation in the transaction of a task from the underlying plan.

wvmcastro: I agreed and copied the marks for all the merges, but then the `creates mission that are fully instanciated` started to fail [buildbot](https://buildbot.tidewise.io/wsgi_dashboards/autoproj/logs/wetpaint:github.com:rock-core:tools-syskit:pulls:443-6/bundles/wetpaint/test). Looks like we can't copy the marks from the toplevel plan tasks to the underlining ones, my guess is that the use cases from engine breaks it but I can't really explain it right now

test/network_generation/test_merge_solver.rb
```diff
@@ -358,6 +358,32 @@ def mock_merged_task_with_concrete_input_connections(*connections)
...
 
+    describe "#apply_merge_group" do
+        attr_reader :local_plan, :solver
+
+        before do
+            @local_plan = Roby::Plan.new
+            @solver = Syskit::NetworkGeneration::MergeSolver.new(@local_plan)
+        end
+
+        # I don't like the "destination task" name but couldn't thought in a better one
```
doudou: Better make that a comment in the PR itself. ... now you have to remove it :P

doudou (approved):

jhonasiv (approved):

PR #444, opened 2024-12-06
feat: capture errors that shouldnt cause the whole plan to be discarded during network resolution

jhonasiv:
Depends on:
- [ ] https://github.com/rock-gazebo/drivers-transformer/pull/13

We introduced a ResolutionError to mark errors during the new plan
network resolution shouldnt be raised, which causes the whole transation
to fail. Instead, we capture them and fail the deployment of the
specific tasks that caused them.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -28,18 +28,12 @@ def minitest_set_options(test_task, name)
     test_task.options = "#{TESTOPTS} #{minitest_args} -- --simplecov-name=#{name}"
 end
 
-def core(early_deploy: false)
-    s = ":no-early-deploy"
-    if early_deploy
-        s = ":early-deploy"
-        early_deploy_setup = ["test/features/early_deploy.rb"]
-    end
-
-    Rake::TestTask.new("test:core#{s}") do |t|
+def core_task(name, files_setup)
+    Rake::TestTask.new("test:core#{name}") do |t|
         t.libs << "."
         t.libs << "lib"
         minitest_set_options(t, "core")
-        test_files = FileList["test/**/test_*.rb", *early_deploy_setup]
+        test_files = FileList["test/**/test_*.rb", *files_setup]
         test_files = test_files
                      .exclude("test/ros/**/*.rb")
                      .exclude("test/gui/**/*.rb")
@@ -48,6 +42,23 @@ def core(early_deploy: false)
         t.test_files = test_files
         t.warning = false
     end
+end
+
+def core(early_deploy: false, capture_errors_during_network_resolution: false,
+    all_features: false)
+    s = ":no-features"
+    if capture_errors_during_network_resolution
+        s = ":capture-errors"
+        files_setup = ["test/features/capture_errors.rb"]
+    elsif early_deploy
+        s = ":early-deploy"
+        files_setup = ["test/features/early_deploy.rb"]
+    elsif all_features
+        s = ":all-features"
+        files_setup = ["test/features/all_features.rb"]
+    end
+
+    core_task(s, files_setup)
 end
 
 Rake::TestTask.new("test:telemetry") do |t|
@@ -78,9 +89,12 @@ Rake::TestTask.new("test:gui") do |t|
 end
 
 core early_deploy: true
+core capture_errors_during_network_resolution: true
+core all_features: true
 core
 desc "Run core library tests, excluding GUI and live tests"
-task "test:core" => ["test:core:no-early-deploy", "test:core:early-deploy"]
+task "test:core" => %w[test:core:no-features test:core:early-deploy
+                       test:core:capture-errors test:core:all-features]
 
 desc "Run all tests"
 task "test" => ["test:gui", "test:core", "test:live", "test:telemetry"]
--- a/lib/syskit/cli/doc/gen.rb
+++ b/lib/syskit/cli/doc/gen.rb
@@ -266,7 +266,7 @@ module Syskit
                 main_plan.add(original_task = model.as_plan)
                 engine = Syskit::NetworkGeneration::Engine.new(main_plan)
                 planning_task = original_task.planning_task
-                mapping = engine.compute_system_network([planning_task], **options)
+                mapping, = engine.compute_system_network([planning_task], **options)
 
                 if engine.work_plan.respond_to?(:commit_transaction)
                     engine.work_plan.commit_transaction
--- a/lib/syskit/exceptions.rb
+++ b/lib/syskit/exceptions.rb
@@ -353,22 +353,20 @@ module Syskit
     # Exception raised when we could not find devices to allocate for tasks
     # that are device drivers
     class DeviceAllocationFailed < SpecError
-        # The set of tasks that failed allocation
-        attr_reader :failed_tasks
+        # The task that failed allocation
+        attr_reader :failed_task
         # A task to parents mapping for tasks involved in this error, at the
         # time of the exception creation
         attr_reader :task_parents
         # Existing candidates for this device
         attr_reader :candidates
 
-        def initialize(plan, tasks)
-            @failed_tasks = tasks.dup
+        def initialize(plan, task)
+            @failed_task = task
             @candidates = {}
             @task_parents = {}
 
-            tasks.each do |abstract_task|
-                resolve_device_task(plan, abstract_task)
-            end
+            resolve_device_task(plan, task)
         end
 
         def resolve_device_task(plan, abstract_task)
@@ -383,7 +381,7 @@ module Syskit
                     all_tasks |= candidates[srv]
                 end
             end
-            self.candidates[abstract_task] = candidates
+            @candidates = candidates
 
             all_tasks.each do |t|
                 next if task_parents.key?(t)
@@ -399,35 +397,33 @@ module Syskit
         end
 
         def pretty_print(pp)
-            pp.text "cannot find a device to tie to #{failed_tasks.size} task(s)"
+            pp.text "cannot find a device to tie to a task"
 
-            failed_tasks.each do |task|
-                parents = task_parents[task]
-                candidates = self.candidates[task]
+            parents = task_parents[failed_task]
+            candidates = self.candidates
+
+            pp.breakable
+            pp.text "for #{failed_task.to_s.gsub('Syskit::', '')}"
+            pp.nest(2) do
+                unless parents.empty?
+                    pp.breakable
+                    pp.seplist(parents) do |parent|
+                        role, parent = parent
+                        pp.text "child #{role.to_a.first} of #{parent.to_s.gsub('Syskit::', '')}"
+                    end
+                end
 
                 pp.breakable
-                pp.text "for #{task.to_s.gsub('Syskit::', '')}"
-                pp.nest(2) do
-                    unless parents.empty?
-                        pp.breakable
-                        pp.seplist(parents) do |parent|
-                            role, parent = parent
-                            pp.text "child #{role.to_a.first} of #{parent.to_s.gsub('Syskit::', '')}"
-                        end
-                    end
-
-                    pp.breakable
-                    pp.seplist(candidates) do |cand|
-                        srv, tasks = *cand
-                        if tasks.empty?
-                            pp.text "no candidates for #{srv.short_name}"
-                        else
-                            pp.text "candidates for #{srv.short_name}"
-                            pp.nest(2) do
-                                pp.breakable
-                                pp.seplist(tasks) do |cand_t|
-                                    pp.text cand_t.to_s
-                                end
+                pp.seplist(candidates) do |cand|
+                    srv, tasks = *cand
+                    if tasks.empty?
+                        pp.text "no candidates for #{srv.short_name}"
+                    else
+                        pp.text "candidates for #{srv.short_name}"
+                        pp.nest(2) do
+                            pp.breakable
+                            pp.seplist(tasks) do |cand_t|
+                                pp.text cand_t.to_s
                             end
                         end
                     end
@@ -467,122 +463,119 @@ module Syskit
     class ConflictingDeploymentAllocation < SpecError
         include Syskit::NetworkGenerationsExceptionHelpers
 
-        attr_reader :deployment_to_tasks
+        attr_reader :orocos_name
 
-        def initialize(deployment_to_tasks, toplevel_tasks_to_requirements = {})
-            @deployment_to_tasks = deployment_to_tasks
+        def initialize(orocos_name, tasks, toplevel_tasks_to_requirements = {})
+            @orocos_name = orocos_name
+            @tasks = tasks
             @toplevel_tasks_to_requirements = toplevel_tasks_to_requirements
-            @deployment_to_execution_agent = \
-                deployment_to_tasks.transform_values do |tasks|
-                    tasks.first.execution_agent
-                end
+            @agent = tasks.first.execution_agent
         end
 
         def pretty_print(pp)
-            deployment_to_tasks.each do |orocos_name, tasks|
-                agent = @deployment_to_execution_agent[orocos_name]
-                deployment_m = agent.deployed_orogen_model_by_name(orocos_name)
-                pp.text(
-                    "deployed task '#{orocos_name}' from deployment " \
-                    "'#{deployment_m.name}' defined in " \
-                    "'#{deployment_m.project.name}' on '#{agent.process_server_name}' " \
-                    "is assigned to #{tasks.size} tasks. Below is the list of " \
-                    "the dependent non-deployed actions. Right after the list " \
-                    "is a detailed explanation of why the first two tasks are not merged:"
+            deployment_m = @agent.deployed_orogen_model_by_name(orocos_name)
+            pp.text(
+                "deployed task '#{orocos_name}' from deployment " \
+                "'#{deployment_m.name}' defined in " \
+                "'#{deployment_m.project.name}' on '#{@agent.process_server_name}' " \
+                "is assigned to #{@tasks.size} tasks. Below is the list of " \
+                "the dependent non-deployed actions. Right after the list " \
+                "is a detailed explanation of why the first two tasks are not merged:"
+            )
+            @tasks.each do |t|
+                defs = find_all_related_syskit_actions(
+                    t, @toplevel_tasks_to_requirements
                 )
-                tasks.each do |t|
-                    defs = find_all_related_syskit_actions(
-                        t, @toplevel_tasks_to_requirements
-                    )
-                    print_dependent_definitions(pp, t, defs)
-                end
-                print_failed_merge_chain(pp, tasks[0], tasks[1])
+                print_dependent_definitions(pp, t, defs)
             end
+            print_failed_merge_chain(pp, @tasks[0], @tasks[1])
         end
     end
 
-    # Exception raised at the end of #resolve if some tasks do not have a
+    # Exception raised at the end of #resolve if a task does not have a
     # deployed equivalent
-    class MissingDeployments < SpecError
-        # The tasks that are not deployed, as a hash from the actual task to
+    class MissingDeployment < SpecError
+        # The tasks that is not deployed, as a hash from the actual task to
         # a set of [role_set, parent_task] pairs
         #
         # This is computed in #initialize as the dependency structure will
         # probably change afterwards
-        attr_reader :tasks
+
+        # The task that is not deployed
+        attr_reader :task
+
+        # Parent tasks of the not deployed task
+        attr_reader :parents
+        # All the available deployment candidates
+        attr_reader :candidates
+        # The deployment hints of the not deployed task
+        attr_reader :deployment_hints
 
         # Initializes this exception by providing a mapping from tasks that
-        # have no deployments to the deployment candidates
+        # have no deployments to the deployment candidates, and the specific task the
+        # error refers to.
         #
         # @param [Hash{TaskContext=>[Array<Model<Deployment>>]}] tasks_with_candidates
-        def initialize(tasks_with_candidates)
-            @tasks = {}
-            tasks_with_candidates.each do |task_to_deploy, candidates|
-                parents = task_to_deploy.dependency_context
-                candidates = candidates.map do |deployed_task, existing_tasks|
-                    existing_tasks = existing_tasks.map do |task|
-                        [task, task.dependency_context]
-                    end
-                    [deployed_task, existing_tasks]
+        def initialize(task, candidates)
+            @task = task
+            @parents = task.dependency_context
+            @candidates = candidates.map do |deployed_task, existing_tasks|
+                existing_tasks = existing_tasks.map do |task|
+                    [task, task.dependency_context]
                 end
-
-                @tasks[task_to_deploy] = [
-                    parents, candidates, task_to_deploy.deployment_hints
-                ]
+                [deployed_task, existing_tasks]
             end
+            @deployment_hints = task.deployment_hints
         end
 
         def pretty_print(pp)
-            pp.text "cannot deploy the following tasks"
-            tasks.each do |task, (parents, possible_deployments)|
+            pp.text "cannot deploy the following task"
+            pp.breakable
+            pp.text "#{task} (#{task.orogen_model.name})"
+            pp.nest(2) do
                 pp.breakable
-                pp.text "#{task} (#{task.orogen_model.name})"
-                pp.nest(2) do
-                    pp.breakable
-                    pp.seplist(parents) do |parent_task|
-                        role, parent_task = parent_task
-                        pp.text "child #{role} of #{parent_task}"
-                    end
+                pp.seplist(parents) do |parent_task|
+                    role, parent_task = parent_task
+                    pp.text "child #{role} of #{parent_task}"
                 end
             end
 
-            tasks.each do |task, (_parents, possible_deployments, deployment_hints)|
-                has_free_deployment = possible_deployments.any? { |_, existing| existing.empty? }
-                pp.breakable
-                if has_free_deployment
-                    pp.text "#{task}: multiple possible deployments, choose one with #prefer_deployed_tasks(deployed_task_name)"
-                    unless deployment_hints.empty?
-                        deployment_hints.each do |hint|
-                            pp.text "  current hints: #{deployment_hints.map(&:to_s).join(', ')}"
-                        end
-                    end
-                elsif possible_deployments.empty?
-                    pp.text "#{task}: no deployments available"
-                else
-                    pp.text "#{task}: some deployments exist, but they are already used in this network"
+            has_free_deployment = candidates.any? { |_, existing| existing.empty? }
+            pp.breakable
+            if has_free_deployment
+                pp.text "#{task}: multiple possible deployments, choose one with " \
+                        "#prefer_deployed_tasks(deployed_task_name)"
+                deployment_hints.each do |hint|
+                    pp.text "  current hints: " \
+                            "#{deployment_hints.map(&:to_s).join(', ')}"
                 end
+            elsif candidates.empty?
+                pp.text "#{task}: no deployments available"
+            else
+                pp.text "#{task}: some deployments exist, but they are already used " \
+                        "in this network"
+            end
 
-                pp.nest(2) do
-                    possible_deployments.each do |deployed_task, existing|
-                        pp.breakable
-                        process_server_name = deployed_task.configured_deployment
-                                                           .process_server_name
-                        orogen_model = deployed_task.configured_deployment
-                                                    .orogen_model
-                        pp.text(
-                            "task #{deployed_task.mapped_task_name} from deployment " \
-                            "#{orogen_model.name} defined in " \
-                            "#{orogen_model.project.name} on #{process_server_name}"
-                        )
-                        pp.nest(2) do
-                            existing.each do |task, parents|
-                                pp.breakable
-                                msg = parents.map do |parent_task|
-                                    role, parent_task = *parent_task
-                                    "child #{role} of #{parent_task}"
-                                end
-                                pp.text "already used by #{task}: #{msg.join(', ')}"
+            pp.nest(2) do
+                candidates.each do |deployed_task, existing|
+                    pp.breakable
+                    process_server_name = deployed_task.configured_deployment
+                                                       .process_server_name
+                    orogen_model = deployed_task.configured_deployment
+                                                .orogen_model
+                    pp.text(
+                        "task #{deployed_task.mapped_task_name} from deployment " \
+                        "#{orogen_model.name} defined in " \
+                        "#{orogen_model.project.name} on #{process_server_name}"
+                    )
+                    pp.nest(2) do
+                        existing.each do |task, parents|
+                            pp.breakable
+                            msg = parents.map do |parent_task|
+                                role, parent_task = *parent_task
+                                "child #{role} of #{parent_task}"
                             end
+                            pp.text "already used by #{task}: #{msg.join(', ')}"
                         end
                     end
                 end
@@ -590,24 +583,25 @@ module Syskit
         end
     end
 
-    # Exception raised at the end of #resolve if some tasks are referring to non-existing
-    # configurations
+    # Exception raised at the end of #resolve if a task is referring to non-existing
+    # configuration
     class MissingConfigurationSection < SpecError
-        # Association of tasks and the sections that are missing
-        attr_reader :missing_sections_by_task
+        # The sections that are missing
+        attr_reader :missing_sections
+        # The task with a missing configuration section
+        attr_reader :task
 
-        def initialize(missing_sections_by_task)
-            @missing_sections_by_task = missing_sections_by_task
+        def initialize(task, missing_sections)
+            @task = task
+            @missing_sections = missing_sections
         end
 
         def pretty_print(pp)
             pp.text "the following configuration sections are used but do not exist"
-            @missing_sections_by_task.each do |task, sections|
-                pp.breakable
-                pp.text "'#{sections.join('\', \'')}', in use by:"
-                pp.breakable
-                pp.text "  #{task} (#{task.orogen_model.name})"
-            end
+            pp.breakable
+            pp.text "'#{missing_sections.join('\', \'')}', in use by:"
+            pp.breakable
+            pp.text "  #{task} (#{task.orogen_model.name})"
         end
     end
 
--- a/lib/syskit/gui/component_network_view.rb
+++ b/lib/syskit/gui/component_network_view.rb
@@ -150,11 +150,11 @@ module Syskit
                 begin
                     if method == :compute_system_network
                         tic = Time.now
-                        @task = compute_system_network(model, plan)
+                        @task, = compute_system_network(model, plan)
                         timing = Time.now - tic
                     elsif method == :compute_deployed_network
                         tic = Time.now
-                        @task = compute_deployed_network(model, plan)
+                        @task, = compute_deployed_network(model, plan)
                         timing = Time.now - tic
                     else
                         @task = instanciate_model(model, plan, instanciate_options)
--- a/lib/syskit/instance_requirements_task.rb
+++ b/lib/syskit/instance_requirements_task.rb
@@ -48,7 +48,6 @@ module Syskit
             root = new_spec.create_proxy_task
             planner = new(**arguments)
             planner.requirements = new_spec
-            root.should_start_after(planner)
             planner.schedule_as(root)
             root.planned_by(planner)
             root
--- a/lib/syskit/instance_requirements_task.rb
+++ b/lib/syskit/instance_requirements_task.rb
@@ -21,6 +21,10 @@ module Syskit
         # @return [InstanceRequirements]
         attr_accessor :requirements
 
+        # Marks the successful end of one network resolution that included this instance
+        # requirement
+        event :resolution_success
+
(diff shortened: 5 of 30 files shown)


lib/syskit/cli/doc/gen.rb
```diff
@@ -72,16 +72,20 @@ def self.save_profile_model(target_path, profile_m)
...
+                    unless resolution_errors.empty?
+                        raise NetworkGeneration::PartialNetworkResolution,
+                              resolution_errors
+                    end
 
                     hierarchy, dataflow =
                         save_profile_definition_graphs(target_path, task, profile_def)
                         .map(&:to_s)
-                rescue StandardError => e
+                rescue StandardError, NetworkGeneration::PartialNetworkResolution => e
```
wvmcastro: `PartialNetworkResolution` is also a `StandardError`, why explicitly capturing it while treating it the same way?

lib/syskit/network_generation/system_network_deployer.rb
```diff
@@ -38,12 +38,14 @@ class SystemNetworkDeployer
             def initialize(plan,
                 event_logger: plan.event_logger,
                 merge_solver: MergeSolver.new(plan),
-                default_deployment_group: Syskit.conf.deployment_group)
+                default_deployment_group: Syskit.conf.deployment_group,
+                required_instances: {})
```
wvmcastro: instances of what ? TaskContext, InstanceRequirements, .. ?

lib/syskit/network_generation/system_network_deployer.rb
```diff
@@ -230,15 +256,25 @@ def verify_all_tasks_deployed
...
-                      "there are tasks for which it exists no deployed equivalent: " \
-                      "#{not_deployed.map { |m| "#{m}(#{m.orogen_model.name})" }}"
+                message = "there are tasks for which it exists no deployed equivalent: " \
+                          "#{not_deployed.map { |m| "#{m}(#{m.orogen_model.name})" }}"
+                tasks_with_candidates.map do |task, candidates|
+                    exception = MissingDeployment.new(task, candidates)
+                    failed_tasks =
+                        self.class.find_instances_from_task(task, required_instances)
+                    failed_tasks.map do |failed|
+                        ResolutionError.new(failed, exception, message)
```
wvmcastro: Here you put the "global" error message into a local resolution problem, I wonder if it won't be confusing or at least noisy

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -353,9 +499,17 @@ def validate_abstract_network
             end
 
             # Validates the network generated by {#compute_system_network}
-            def validate_generated_network
-                self.class.verify_task_allocation(plan)
-                self.class.verify_device_allocation(plan, toplevel_tasks_to_requirements)
+            def validate_generated_network(toplevel_tasks: @toplevel_tasks,
+                merge_solver: @merge_solver)
```
wvmcastro: some plugin will pass its own toplevel_tasks or merge_solver?

wvmcastro (commented):
One thing that is not clear to me, when trying to deploy the below composition there is an error with second deployment and `on_error: commit`. The deployer will still deploy first and second child?
```ruby
class Parent < Compositions
  add Model1, as: "first"
  add Model2, as "second"
  add Model3, as "third"
  ...
end
```

lib/syskit/network_generation/system_network_generator.rb
jhonasiv: some unit tests do

lib/syskit/network_generation/system_network_deployer.rb
jhonasiv: I suppose I could make this message individual for each task without much hassle

lib/syskit/network_generation/system_network_deployer.rb
jhonasiv: This parameter is a bit too verbose to be too explicit, I was actually hoping to find another will of passing it through instead of making it a member, unfortunately I cant see a way without changing the interface of #deploy.

jhonasiv: I will add some documentation nonetheless

lib/syskit/cli/doc/gen.rb
jhonasiv: Thats an oversight on my part

lib/syskit/network_generation/system_network_generator.rb
jhonasiv: Nope, any plugin should implement this version of the method themselves (that's where the super comes in). I will resolve these arguments, they arent needed.

jhonasiv:
> One thing that is not clear to me, when trying to deploy the below composition there is an error with second deployment and `on_error: commit`. The deployer will still deploy first and second child?
> 
> ```ruby
> class Parent < Compositions
>   add Model1, as: "first"
>   add Model2, as "second"
>   add Model3, as "third"
>   ...
> end
> ```

If a child of a parent fails the deployment validation, itself and all of its parents also fail, therefore the other children will also fail. PS: `on_error: commit` is only used by profile_assertions tests, so in normal circumstances the deployer will do what I just said, but when on_error: commit the whole transaction would be commited, which would lead to all present errors being propagated at the end of the network generation process.

lib/syskit/network_generation/system_network_deployer.rb
jhonasiv: Added documentation only for the required instances parameter, as the others are self-explanatory AND already explained in the attr_readers

test/test/test_network_manipulation.rb
```diff
@@ -376,7 +387,7 @@ module Test
                 end
 
                 it "fails if the plan can't be deployed" do
-                    plan.add_mission_task(@cmp_m.as_plan)
+                    plan.add_mission_task(@cmp_m.use("srv" => @task_m).as_plan)
```
jhonasiv: I dont remember adding this line...Anyways, this is failing, somehow instead of UnexpectedErrors I am getting the actual error TaskAllocationFailed. Same thing if I remove this specific change. Still looking into it....

jhonasiv: This was an issue with the return value of apply_requirements_modifications#syskit_apply_async_resolution_results. I had to update it to be more informative instead of reporting either nil or error. Doing a full pass in the unit tests right now, then I will push this change

Rakefile
```diff
@@ -78,9 +81,11 @@ Rake::TestTask.new("test:gui") do |t|
 end
 
 core early_deploy: true
+core capture_errors_during_network_resolution: true
```
doudou, line 92: I think we should run with both activated as well ... no ?

lib/syskit/network_generation/engine.rb
```diff
@@ -701,32 +709,44 @@ def self.discover_requirement_tasks_from_plan(plan)
...
                     Engine.discover_requirement_tasks_from_plan(real_plan),
+                error_handler: RaiseErrorHandler.new,
                 garbage_collect: true,
                 validate_abstract_network: true,
                 validate_generated_network: true,
                 default_deployment_group: Syskit.conf.deployment_group,
-                validate_deployed_network: (true if Syskit.conf.early_deploy?),
-                early_deploy: Syskit.conf.early_deploy?
+                validate_deployed_network: Syskit.conf.early_deploy?,
+                early_deploy: false,
```
doudou: Why the change from `Syskit.conf.early_deploy?`, especially given that you kept it for `validate_deployed_network`

lib/syskit/network_generation/engine.rb
```diff
@@ -916,6 +962,21 @@ def commit_work_plan
                 end
             end
 
+            def handle_resolution_errors(errors,  on_error: :discard)
+                return if work_plan.finalized? || work_plan == real_plan
+
+                if on_error == :save
+                    errors.each do |e|
+                        handle_resolution_exception(e, on_error: on_error)
```
doudou: It really does not make sense to call this more than once.

Also, the (very little) difference in logic between handle_resolution_errors and handle_resolution_exception looks suspicious.

lib/syskit/network_generation/engine.rb
```diff
@@ -831,15 +868,24 @@ def resolve(
...
                     compute_policies: compute_policies,
                     validate_deployed_network: validate_deployed_network,
-                    early_deploy: early_deploy
+                    early_deploy: early_deploy,
+                    capture_errors_during_network_resolution:
+                        capture_errors_during_network_resolution,
+                    cleanup_resolution_errors: cleanup_resolution_errors
                 )
 
+                unless resolution_errors.empty? || cleanup_resolution_errors
```
doudou: I find that `unless(complex logic)` is very hard to read ... I'd keep it for simple logic (only one boolean)

Here, `if !resolution_errors.empty? && !cleanup_resolution_errors`

lib/syskit/network_generation/engine.rb
```diff
@@ -916,6 +962,21 @@ def commit_work_plan
...
+                return if work_plan.finalized? || work_plan == real_plan
+
+                if on_error == :save
+                    errors.each do |e|
+                        handle_resolution_exception(e, on_error: on_error)
+                    end
+                elsif on_error == :commit
+                    work_plan.commit_transaction
+                else
+                    errors.each { |err| real_plan.execution_engine.add_error(err) }
```
doudou: I'm not very happy that you modify `real_plan` from here

lib/syskit/network_generation/system_network_deployer.rb
```diff
@@ -244,15 +248,21 @@ def self.verify_all_tasks_deployed(plan, default_deployment_group)
...
-                      "there are tasks for which it exists no deployed equivalent: " \
-                      "#{not_deployed.map { |m| "#{m}(#{m.orogen_model.name})" }}"
+                tasks_with_candidates.each do |task, candidates|
+                    message =
+                        "#{task}(#{task.orogen_model.name}) has no deployed equivalent"
+                    exception = MissingDeployment.new(task, candidates)
+                    error_handler.register_resolution_failures_from_exception(
+                        [task], exception, message
+                    )
+                end.flatten
```
doudou: I think the flatten and the `return []` above are leftovers from the previous error reportnig mechanism

lib/syskit/network_generation/system_network_deployer.rb
```diff
@@ -271,10 +281,15 @@ def verify_all_configurations_exist
...
 
-                raise MissingConfigurationSection.new(missing),
-                      "some configuration sections are used but not defined"
+                message = "some configuration sections are used but not defined"
+                missing.map do |task, sections|
+                    exception = MissingConfigurationSection.new(task, sections)
+                    error_handler.register_resolution_failures_from_exception(
+                        [task], exception, message
+                    )
+                end.flatten
```
doudou: Same

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -397,41 +481,54 @@ def self.verify_all_deployments_are_unique(
                     tasks.size > 1
                 end
 
-                return if using_same_deployment.empty?
+                return [] if using_same_deployment.empty?
```
doudou: Same ... unnecessary ?

lib/syskit/network_generation/async.rb
```diff
@@ -133,14 +133,17 @@ def apply
                 engine = future.engine
                 if @cancelled
                     engine.discard_work_plan
-                    false
+                    SystemNetworkPlanApplyResult.new(fulfilled: false)
```
doudou: Why not keep `nil` here to indicate that the application is cancelled (and remove the `fulfilled` flag.

It's not that the future is fulfilled or not, it's that the resolution got cancelled (so, we throw it away)

lib/syskit/network_generation/engine.rb
```diff
@@ -819,10 +853,13 @@ def resolve(
                 validate_generated_network: true,
                 validate_deployed_network: true,
                 validate_final_network: true,
-                early_deploy: Syskit.conf.early_deploy?
+                early_deploy: Syskit.conf.early_deploy?,
+                capture_errors_during_network_resolution:
+                    Syskit.conf.capture_errors_during_network_resolution?,
+                cleanup_resolution_errors: on_error != :commit
```
doudou, line 857: I don't understand that default. Maybe just capture_errors_during_network_resolution ?

lib/syskit/runtime/apply_requirement_modifications.rb
```diff
@@ -102,21 +102,30 @@ def syskit_apply_async_resolution_results
...
+                    task.failed_event.emit e.flat_map(&:original_exception)
+                end
+                resolution_apply_result
             rescue ::Exception => e # rubocop:disable Lint/RescueException
                 running_requirement_tasks.each do |t|
                     t.failed_event.emit(e)
                 end
-                e
+                NetworkGeneration::SystemNetworkPlanApplyResult.new(
+                    fulfilled: false, errors: [e]
```
doudou: Definitely need to get rid of `fulfilled`, which have different meanings depending on the context.

lib/syskit/test/instance_requirement_planning_handler.rb
```diff
@@ -56,17 +56,17 @@ def apply_requirements
...
+            def finished? # rubocop:disable Metrics/AbcSize
                 Thread.pass
 
                 if @plan.syskit_has_async_resolution?
                     return unless @plan.syskit_finished_async_resolution?
 
-                    error = @plan.syskit_apply_async_resolution_results
-                    return true if error
+                    resolution_results = @plan.syskit_apply_async_resolution_results
+                    return true unless resolution_results.fulfilled?
```
doudou: The special error case is not needed anymore, one can go on, find out that there are e.g. no root tasks (because the whole resolution failed) and then get into `@planning_tasks.all?(&:finished?)`

lib/syskit/test/network_manipulation.rb
```diff
@@ -160,11 +161,21 @@ def syskit_generate_network(*to_instanciate, add_missions: true)
...
                         validate_generated_network: false,
                         early_deploy: false
                     )
+                    unless resolution_errors.empty?
+                        resolution_errors.each do |error|
+                            t = error.planned_task
+                            expect_execution do
+                                t.failed_event.emit(error.original_exceptions)
+                            end
+                        end
```
doudou: Would be better to invert, that is run the loop inside the expect_execution

And replace expect_execution by execute. expect_execution without a `to` block actually does nothing

lib/syskit/test/network_manipulation.rb
```diff
@@ -227,31 +238,32 @@ def syskit_deploy_normalized_placeholder_tasks(
                     .to { emit(*not_running.map(&:start_event)) }
 
                 resolve_options = Hash[on_error: :commit].merge(resolve_options)
-                begin
+                resolution_errors = execute do
```
doudou: Why the need for `execute` here ?

lib/syskit/test/network_manipulation.rb
```diff
@@ -227,31 +238,32 @@ def syskit_deploy_normalized_placeholder_tasks(
...
-                end
-
-                execute do
-                    placeholder_tasks.each do |task|
-                        plan.remove_task(task)
+                else
+                    resolution_errors.each do |error|
+                        t = error.planned_task
+                        expect_execution do
+                            t.failed_event.emit(error.original_exception)
```
doudou, line 263: You've got a problem here if the task was from a previous deploy (can't emit failed_event)

test/network_generation/test_system_network_generator.rb
```diff
@@ -210,16 +210,268 @@ def arg=(value)
...
+                              #<Class:0xXXXXXX>.use(t1 => T<id:X>(arg: 1, conf: ["default"], read_only: false, test_dev: device(D, as: test)), t2 => T<id:X>(arg: 2, conf: ["default"], read_only: false, test_dev: device(D, as: test)))
+                            T<id:X>(arg: 1, conf: ["default"], read_only: false, test_dev: device(D, as: test)) is needed by the following definitions:
+                              #<Class:0xXXXXXX>.use(t1 => T<id:X>(arg: 1, conf: ["default"], read_only: false, test_dev: device(D, as: test)), t2 => T<id:X>(arg: 2, conf: ["default"], read_only: false, test_dev: device(D, as: test)))
+                        PP
+                        errors.each do |err|
+                            assert_exception(err, requirement_tasks.first.planned_task,
+                                             expected_message)
+                            # formatted = formatted.gsub(/<id:\d+>/, "<id:X>")
+                            #                      .gsub(/Class:0x[0-9a-f]+/,
+                            #                            "Class:0xXXXXXX")
```
doudou: leftover

test/network_generation/test_system_network_generator.rb
```diff
@@ -210,16 +210,268 @@ def arg=(value)
...
+                            Models::Placeholder<#<#<Class:0xXXXXXX>:0xXXXXXX>><id:X>()
+                              no candidates
+                              child task of <id:X>(conf: [])
+                        MSG
+                        assert_exception(error, requirement_tasks.first.planned_task,
+                                         expected_message)
+                        # assert_equal expected,
+                        #              formatted.gsub(/<id:\d+>/, "<id:X>")
+                        #                       .gsub(/Class:0x[0-9a-f]+>:0x[0-9a-f]+>/,
+                        #                             "Class:0xXXXXXX>:0xXXXXXX>")
```
doudou: leftover

lib/syskit/runtime/apply_requirement_modifications.rb
```diff
@@ -102,21 +102,30 @@ def syskit_apply_async_resolution_results
...
                     @syskit_current_resolution = nil
                 end
 
-                running_requirement_tasks.each do |t|
+                resolution_apply_result.instances.each_key do |t|
                     t.success_event.emit
                 end
-                nil
+                resolution_apply_result.errors.group_by(&:planning_task).each do |task, e|
+                    task.failed_event.emit e.flat_map(&:original_exception)
```
doudou: `task` is not guaranteed to be running here. Might be from a previous deploy.

lib/syskit/network_generation/engine.rb
doudou: Would be better to make early_deploy the default for validate_deployed_network

```
early_deploy: Syskit.conf.early_deploy?,
validate_deployed_network: early_deploy
```

lib/syskit/runtime/apply_requirement_modifications.rb
```diff
@@ -102,21 +102,30 @@ def syskit_apply_async_resolution_results
...
+                    unless resolution_apply_result.fulfilled?
+                        return resolution_apply_result
+                    end
                 ensure
                     syskit_current_resolution_keepalive.discard_transaction
                     @syskit_current_resolution = nil
                 end
 
-                running_requirement_tasks.each do |t|
+                resolution_apply_result.instances.each_key do |t|
```
doudou: Instead of finishing InstanceRequirementsTask when the resolution finishes, my proposition would be to

- have an intermediate event `resolution_success`, which gets emitted here instead of `success`
- change the logic that finds what needs to be deployed to only pick the running InstanceRequirementsTask
- change the logic that finds whether there are new requirements to check if there are running InstanceRequirementsTask without the resolution_success event emitted (resolution_success_event.emitted?)

doudou: Logic in runtime/apply_deployment_tasks_from_plan.rb

lib/syskit/runtime/apply_requirement_modifications.rb
doudou: You should be returning a SystemNetworkPlanApplyResult that has the same instances as before the resolution, and a proper list of errors attached to the new ones.

lib/syskit/test/network_manipulation.rb
```diff
@@ -160,11 +161,21 @@ def syskit_generate_network(*to_instanciate, add_missions: true)
...
                     engine = NetworkGeneration::Engine.new(plan, work_plan: trsc)
-                    mapping = engine.compute_system_network(
+                    mapping, resolution_errors = engine.compute_system_network(
                         tasks_to_instanciate.map(&:planning_task),
                         validate_generated_network: false,
                         early_deploy: false
                     )
+                    unless resolution_errors.empty?
+                        resolution_errors.each do |error|
+                            t = error.planned_task
```
doudou: `planning_task` ?

lib/syskit/network_generation/engine.rb
doudou: Conclusion: should be removed.

test/features/capture_errors.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Even though the early deploy feature isnt a direct requirement to the error capture,
+# it adds more points where exception might be raised. So it makes sense to leave it
+# enabled
+Syskit.conf.early_deploy = true
+Syskit.conf.capture_errors_during_network_resolution = true
```
doudou: Decouple the two flags, and run 4 test suites (plain core, with early deploy, with capture errors and with both)

lib/syskit/network_generation/engine.rb
jhonasiv: The idea here is that if we clean up the plan with whatever resolution errors we get the committed transaction wouldnt include the failing instance requirement tasks, therefore it wouldnt raise errors in the same manner as before, which matter for contexts that expect them (i.e. network_manipulation and syskit/gui)

lib/syskit/gui/component_network_view.rb
```diff
@@ -150,11 +150,17 @@ def render(model,
                 begin
                     if method == :compute_system_network
                         tic = Time.now
-                        @task = compute_system_network(model, plan)
+                        @task, resolution_errors = compute_system_network(model, plan)
+                        exception = NetworkGeneration::PartialNetworkResolution.new(
+                            resolution_errors
+                        )
```
doudou: Isn't that reporting an error in all cases ? Even when the deploy was fine ?

lib/syskit/instance_requirements_task.rb
```diff
@@ -21,6 +21,9 @@ class InstanceRequirementsTask < Roby::Task
         # @return [InstanceRequirements]
         attr_accessor :requirements
 
+        # Marks the end of the network resolution of this instance requirement task.
```
doudou: Marks the successful end of one network resolution that included this instance requirement

lib/syskit/network_generation/engine.rb
```diff
@@ -831,15 +861,25 @@ def resolve(
...
                     compute_policies: compute_policies,
                     validate_deployed_network: validate_deployed_network,
-                    early_deploy: early_deploy
+                    early_deploy: early_deploy,
+                    capture_errors_during_network_resolution:
+                        capture_errors_during_network_resolution,
+                    cleanup_resolution_errors: cleanup_resolution_errors
                 )
 
+                if !resolution_errors.empty? && !cleanup_resolution_errors
```
doudou, line 877: I'd like to at least have a comment here that this piece of code won't be reached if capture_errors_during_network_resolution is false

lib/syskit/network_generation/system_network_deployer.rb
```diff
@@ -204,32 +204,36 @@ def apply_selected_deployments(selected_deployments, deployment_tasks = {})
...
                                    .not_finished.not_abstract
                                    .find_all { |t| !t.execution_agent }
 
                 return if not_deployed.empty?
 
                 tasks_with_candidates = {}
-                not_deployed.each do |task|
+                # This is reversed to give preference to child tasks, as they contain
+                # more information about the dependency context then their parents
+                not_deployed.reverse_each do |task|
```
doudou: not_deployed comes from find_local_tasks which (1) has no ordering guarantees whatsoever (probably is ordered by creation time) and (2) is made of TaskContext, which in the very very common case are tasks that have no children.

lib/syskit/network_generation/engine.rb
```diff
@@ -681,52 +689,57 @@ def adapt_existing_deployment(deployment_task, existing_deployment_task)
...
-                needed = plan.useful_tasks(with_transactions: false)
-                req_tasks.delete_if do |t|
-                    !needed.include?(t)
-                end
-                req_tasks
+                req_tasks =
+                    plan.find_local_tasks(InstanceRequirementsTask).running
+                req_tasks = req_tasks.find_all do |t|
+                    planned_task = t.planned_task
+                    !planned_task&.finished? || planned_task&.being_repaired?
```
doudou: `!planned_task&.finished?` == true if planned_task.nil?

Not sure that's really what we want...

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -399,39 +483,52 @@ def self.verify_all_deployments_are_unique(
...
+                end
             end
 
             # Validates the network generated by {#compute_system_network}
             #
             # It performs the tests that are only needed on an abstract network,
             # i.e. on a network in which some tasks are still abstract
-            def validate_abstract_network
+            def validate_abstract_network(error_handler: @error_handler)
                 self.class.verify_no_multiplexing_connections(plan)
```
doudou, line 505: Not passing error_handler to that one ?

lib/syskit/test/instance_requirement_planning_handler.rb
```diff
@@ -56,17 +56,21 @@ def apply_requirements
                 )
             end
 
-            def finished?
+            def finished? # rubocop:disable Metrics/AbcSize
```
doudou: Did we talk about disabling rubocop cops ?

lib/syskit/test/instance_requirement_planning_handler.rb
```diff
@@ -83,7 +87,7 @@ def finished?
                     stub_network.announce_replacements(mapped_tasks)
                     stub_network.remove_obsolete_tasks(mapped_tasks)
                 end
-                @planning_tasks.all?(&:finished?)
+                @planning_tasks.all? { |t| t.resolution_success? || t.failed? }
```
doudou: I would use `t.finished?` instead of `t.failed?`. It's more general and would cover all cases (even though I'm ... pretty sure ... that the only way to quit is to fail.

lib/syskit/test/network_manipulation.rb
```diff
@@ -173,7 +188,9 @@ def syskit_generate_network(*to_instanciate, add_missions: true)
                         replacement = task_mapping[task.planning_task]
                         plan.replace_task(task, replacement)
                         plan.remove_task(task)
-                        replacement.planning_task.success_event.emit
+                        # This breaks repaired tasks tests, but without it, a whole lot of stub tests also break?!
+                        # replacement.planning_task.success_event.emit
```
doudou: Still relevant ?

lib/syskit/gui/component_network_view.rb
jhonasiv: Oops

lib/syskit/test/instance_requirement_planning_handler.rb
jhonasiv: I did see this and was planning to resolve it but forgot...thanks for the reminder

lib/syskit/test/network_manipulation.rb
jhonasiv: Nop, just forgot to remove the comment

lib/syskit/network_generation/system_network_generator.rb
jhonasiv: No because it is one of the sanity checks we talked about that should indeed raise as it indicates an error in the plan resolution

lib/syskit/gui/component_network_view.rb
jhonasiv: Ah, I know what happened here, this code is a remnant from the last iteration (the one where we didnt have the ErrorHandler concept), and I failed to remove it entirely. In the current iteration, this will retain the original behavior, aka it will raise, so I can this remove any resolution_errors mention.

lib/syskit/network_generation/system_network_deployer.rb
jhonasiv: This is a lazy solution to be honest, it was supposed to leave unit tests undisturbed but its clearly a dirty workaround. I will do the proper fixes instead.

jhonasiv: I ended up sorting the deployments by the number of parents, so this is done explicitly instead of relying on reversing the not deployed list. The old behavior of this part is a bit tricky to maintain as it raised for all the not deployed tasks before, while now I need to raise for a single one. This change targets a single unit test that checked for the parent in test_system_network_deployer (the test name is snapshots the task's parents at the exception point).

test/features/all_features.rb
wvmcastro, line 1: maybe just require each feature file? Thinking if it could be automated to require all files at this dir except for this one (no need to do it right now, just thinking outloud)

lib/syskit/exceptions.rb
```diff
@@ -467,147 +463,147 @@ def pretty_print(pp)
...
-                        end
+            has_free_deployment = candidates.any? { |_, existing| existing.empty? }
+            pp.breakable
+            if has_free_deployment
+                pp.text "#{task}: multiple possible deployments, choose one with " \
+                        "#prefer_deployed_tasks(deployed_task_name)"
+                unless deployment_hints.empty?
+                    deployment_hints.each do |hint|
+                        pp.text "  current hints: " \
+                                "#{deployment_hints.map(&:to_s).join(', ')}"
```
wvmcastro: Since this is already inside `#each` block, I don't understand the `unless deployment_hints.empty?` check

test/network_generation/test_system_network_deployer.rb
```diff
@@ -463,19 +463,20 @@ def make_candidates(count)
...
                             deployer.validate_deployed_network
                         end
 
-                        info = e.tasks[task]
-                        assert_equal [], info[0] # parents
-                        candidates = info[1]
+                        parents = e.parents
+                        assert_equal [], parents
+                        candidates = e.candidates
+                        hints = e.deployment_hints
```
wvmcastro, line 474: This is so much more readable o/

lib/syskit/network_generation/resolution_error_handler.rb
```diff
@@ -0,0 +1,229 @@
...
+            )
+                return [] unless toplevel_tasks
+
+                # Build the mapping of the toplevel task into the actual task
+                # in the plan that represents it
+                toplevel_tasks_indexes =
+                    toplevel_tasks
+                    .each_with_index
+                    .group_by { |t, _i| merge_solver.replacement_for(t) }
+                    .transform_values { |v| v.flatten[1] }
```
wvmcastro, line 130: I didn't get why always taking the second top level task as the dependent one. Self reminder to come back at this later

wvmcastro (commented):
As we talked at the office, I am submitting this review but I am not done yet

lib/syskit/exceptions.rb
jhonasiv: Oh, I didnt see that when changing the message, its indeed not necessary

lib/syskit/network_generation/resolution_error_handler.rb
jhonasiv: The `group_by` returns a Hash that has the merge solver replacement as keys and multiple [toplevel_task, index] as values. The `transform_values` call takes the indexes as all I need past this point a mapping between the toplevel task replacement and the indexes of the toplevel tasks that were replaced

test/features/all_features.rb
jhonasiv: Makes sense, though I dont think it warrants an automation for such a simple code, any time one adds a new feature flag they simply need to add a single include in this file.

lib/syskit/network_generation/resolution_error_handler.rb
wvmcastro: Right! I didn't realize the values where `[toplevel_task, index]`. Tks

lib/syskit/network_generation/engine.rb
```diff
@@ -105,7 +108,12 @@ def compute_deployed_network(
                         default_deployment_group: default_deployment_group
                     )
 
-                    deployer.deploy(validate: validate_deployed_network)
+                    deployer.deploy(
+                        error_handler: error_handler, validate: validate_deployed_network
+                    )
+                    resolution_errors = error_handler.process_failures(
+                        required_instances, cleanup_failed_tasks: true
```
wvmcastro, line 115: The `cleanup_failed_tasks` toggle is really used for toggling? Looks like it is always true

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -216,37 +222,40 @@ def deploy(deployment_tasks)
...
             end
 
+            def instanciate_system_network(instance_requirements)
+                @toplevel_tasks = log_timepoint_group "instanciate" do
+                    instanciate(instance_requirements)
+                end
+                Engine.instanciation_postprocessing.each do |block|
+                    block.call(self, plan)
+                    log_timepoint "postprocessing:#{block}"
+                end
```
wvmcastro, line 232: Why instantiation postprocessing is being calles twice, here and after deploy?

test/network_generation/test_system_network_generator.rb
```diff
@@ -220,6 +220,258 @@ def arg=(value)
...
+                                garbage_collect: garbage_collect
+                            )
+                            required_instances =
+                                Hash[requirement_tasks.zip(toplevel_tasks)]
+                            errors = error_handler.process_failures(
+                                required_instances, cleanup_failed_tasks: true
+                            )
+                            [toplevel_tasks, errors]
+                        end
+                    end
```
wvmcastro, line 446: I think here you intend to test the flow _insciate_system_network -> resolve_system_network -> process_failures_, the same used at `Engine#compute_system_network`. Shouldn't it be a method and write the test for that method instead?

lib/syskit/network_generation/system_network_generator.rb
doudou: It is 'instanciated_network_postprocessing' after deploy

lib/syskit/network_generation/system_network_deployer.rb
jhonasiv: Still the wrong fix, @doudou figured out a change to unit test

lib/syskit/network_generation/engine.rb
jhonasiv: It is, there is this code path where it can be toggled (and by default it is false when on_error == :commit):

1. Engine#resolve, that forwards cleanup_failed_tasks to
2. Engine#resolve_system_network, that that forwards cleanup_failed_tasks to
3. Engine#compute_system_network, that actually cleans it up

In the part of the code you referenced, we always want to cleanup.

lib/syskit/network_generation/system_network_generator.rb
jhonasiv: You are getting confused, in here its `instanciation_postprocessing`, the one after deploy is `instanciated_network_postprocessing`

lib/syskit/network_generation/system_network_deployer.rb
```diff
@@ -244,15 +250,29 @@ def self.verify_all_tasks_deployed(plan, default_deployment_group)
...
+                # about the error.
+                sort_exceptions =
+                    exceptions.sort do |a, b|
+                        b.parents.size <=> a.parents.size
+                    end
+                sort_exceptions.each do |e|
+                    error_handler.register_resolution_failures_from_exception(
+                        [e.task], e, e.message
+                    )
+                end
```
doudou: So, for the record, remove the sort.

lib/syskit/network_generation/system_network_generator.rb
doudou: For the record: neither the composition model creation, nor the merge solver, should allow the creation of multiplexing connections. So, indeed, sanity checl.

test/network_generation/test_system_network_generator.rb
jhonasiv: Engine has its own tests, in here I wanted to test that I am capturing specific errors in the SystemNetworkGenerator itself, to do that I had to replicate the flow from Engine#compute_system_network, but I dont need any additional logic, which the Engine has.

doudou (approved):
Only the sort missing... I'm preemptively approving.

lib/syskit/network_generation/system_network_generator.rb
wvmcastro: Yes, I got confused

wvmcastro (approved):

lib/syskit/runtime/apply_requirement_modifications.rb
```diff
@@ -102,21 +102,29 @@ def syskit_apply_async_resolution_results
...
+
+                    return unless resolution_apply_result
                 ensure
                     syskit_current_resolution_keepalive.discard_transaction
                     @syskit_current_resolution = nil
                 end
 
-                running_requirement_tasks.each do |t|
-                    t.success_event.emit
+                resolution_apply_result.instance_requirement_tasks.each do |t|
```
wvmcastro, line 113: What would be the difference between `running_requirement_tasks` and `resolution_apply_result.instance_requirement_tasks` ?

jhonasiv: instance requirement tasks are the requirement tasks that were well defined and therefore their deployments were successful. `running_requirement_tasks` are all the req tasks that are considered for planning

PR #445, opened 2024-12-06
feat: early deploy at network generation

wvmcastro:
On top of:
- [ ] https://github.com/rock-core/tools-syskit/pull/443

Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/304

Opening this PR for getting a CI run with test failures and an early request for comments


lib/syskit/exceptions.rb
```diff
@@ -483,6 +483,39 @@ def pretty_print(pp)
         end
     end
 
+    class ConflictingDeploymentAllocationAtNetworkGeneration < SpecError
```
doudou: The `AtNetworkGeneration` is not meaningful. I don't care where this is detected, only that it is.

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -254,6 +297,17 @@ def compute_system_network(instance_requirements, garbage_collect: true,
                 @toplevel_tasks
             end
 
+            def merge_identical_tasks
+                # When early deploying MergeSolver must merge tasks with identical
+                # execution agents. This breaks a strong assumption for the solver that is
+                # if two tasks have execution agents, then they must be different. But
+                # when early deploying tasks that have the same agent are actually the
+                # same
+                merge_solver.merge_when_identical_agents = early_deploy?
```
doudou: Do you have to un-set it ? Can't it be set to early_deploy? once and for all ?

test/test/test_profile_assertions.rb
```diff
@@ -721,6 +680,8 @@ module Test
                 end
 
                 it "allows deploying together with the actions or profile" do
+                    skip if Syskit.conf.early_deploy?
+
```
doudou: Why the `skip` ? Please add a comment with the reason ...

test/test/test_profile_assertions.rb
```diff
@@ -668,49 +670,6 @@ module Test
...
-
-                    e = assert_raises(ProfileAssertions::ProfileAssertionFailed) do
-                        assert_can_deploy_all(@test_profile)
-                    end
-                    assert_match(
-                        /cannot find a concrete implementation.*Other.test_tag/m,
-                        PP.pp(e.each_original_exception.first, +"")
-                    )
-                end
-
```
doudou: You're removing tests, but I don't see any new tests that would cover the same functionality...

wvmcastro: These same testes have counter parts at _assert_can_instanciate_ section (with the same test name also), and I think they are actually instantiation tests because don't having a concrete implementation for a placeholder is primitive error at network generation

lib/syskit/network_generation/system_network_generator.rb
wvmcastro: This is the engine merge solver, some engine tests fail when I let this set once and for all. I can investigate more on what engine assumption about the merge solver it breaks

test/test/test_profile_assertions.rb
wvmcastro: I am actually struggling to remember the reason .. what I know is that when deploying the composition child the deployer does not find a deployed task at the deployment group later when verifying that all tasks are deployed we get "there are tasks for which it exists no deployed equivalent". What I can't explain is why the deployment group of this test's deployment model is different from the one passed to engine

lib/syskit/exceptions.rb
```diff
@@ -483,6 +483,39 @@ def pretty_print(pp)
...
+                end
+                pp.breakable
+                pp.text "because the same "
+                process_server_name = deployed_task.configured_deployment
+                                                   .process_server_name
+                orogen_model = deployed_task.configured_deployment
+                                            .orogen_model
+                pp.text(
+                    "deployed task #{deployed_task.mapped_task_name} from deployment " \
+                    "#{orogen_model.name} defined in " \
```
jhonasiv:
```suggestion
                    "#{orogen_model.name} is defined in " \
```

lib/syskit/network_generation/merge_solver.rb
```diff
@@ -220,16 +218,39 @@ def may_merge_task_contexts?(merged_task, task)
...
+
+            # Tests whether task.merge(target_task) is a valid operation
+            #
+            # @param [Syskit::TaskContext] task
+            # @param [Syskit::TaskContext] target_task
+            #
+            # @return [false,true] if false, the merge is not possible. If
+            #   true, it is possible. If nil, the only thing that makes the
+            #   merge impossible are missing inputs, and these tasks might
+            #   therefore be merged if there was a dataflow cycle
```
jhonasiv: nil is not a possibility

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -254,6 +299,17 @@ def compute_system_network(instance_requirements, garbage_collect: true,
...
 
+            def merge_identical_tasks
+                # When early deploying MergeSolver must merge tasks with identical
+                # execution agents. This breaks a strong assumption for the solver that is
+                # if two tasks have execution agents, then they must be different. But
+                # when early deploying tasks that have the same agent are actually the
+                # same
+                merge_solver.merge_when_identical_agents = deploy_config.early_deploy?
+                merge_solver.merge_identical_tasks
+                merge_solver.merge_when_identical_agents = false
```
jhonasiv: A better design is to create a block-context when dealing with `merge_when_identical_agents` (something I can actually use in the error-capture PR I am working on as well). It would something like a python **with** statement, e.g.
```
merge_solver.allow_identical_agents do
    merge_solver.merge_identical_agents
end
```

Make that the only way of allowing identical agents, which guarantees that no merge_when_identical_agents = true is left without its false counterpart.

test/test/test_profile_assertions.rb
wvmcastro: We did the task allocation verification before the deployment

lib/syskit/network_generation/merge_solver.rb
wvmcastro: The doc seems to be deprecated already, before this change. Tks

lib/syskit/network_generation/system_network_generator.rb
wvmcastro: Went with Jhonas' suggestion https://github.com/rock-core/tools-syskit/pull/445#discussion_r1940006731

wvmcastro:
ping @doudou @jhonasiv

lib/syskit/exceptions.rb
```diff
@@ -483,6 +483,39 @@ def pretty_print(pp)
         end
     end
 
+    class ConflictingDeploymentAllocation < SpecError
```
doudou, line 467: One good thing is to explicitly test the pretty_print method.

You can do it by creating the exception object and then

`text = PP.pp(exception, +"")`

Afterwards there's usually some sanitizing to do (`id:\d+` and `0x[0-9a-f]+` need to be removed because they vary from call to call)

lib/syskit/network_generation/engine.rb
```diff
@@ -703,19 +703,31 @@ def compute_system_network(
...
-                    instance_requirements,
-                    garbage_collect: garbage_collect,
-                    validate_abstract_network: validate_abstract_network,
-                    validate_generated_network: validate_generated_network
+                    work_plan,
+                    event_logger: event_logger,
+                    merge_solver: merge_solver,
+                    default_deployment_group: default_deployment_group,
+                    early_deploy: early_deploy,
+                    validate_deployed_network: validate_deployed_network
```
doudou, line 726: I don't think it makes sense to have `validate_deployed_network` here if all the `validate_*` options are passed to `generate`

lib/syskit/network_generation/engine.rb
```diff
@@ -703,19 +703,31 @@ def compute_system_network(
                     Engine.discover_requirement_tasks_from_plan(real_plan),
                 garbage_collect: true,
                 validate_abstract_network: true,
-                validate_generated_network: true
+                validate_generated_network: true,
+                default_deployment_group: nil,
+                validate_deployed_network: false,
+                early_deploy: false
```
doudou: Shouldn't the default be `Syskit.conf.early_deploy?`

lib/syskit/network_generation/merge_solver.rb
```diff
@@ -220,16 +216,49 @@ def may_merge_task_contexts?(merged_task, task)
...
+                true
+            end
+
+            def merge_task_contexts_with_same_agent(enabled:)
+                @merge_when_identical_agents = enabled
+                r = yield
+                @merge_when_identical_agents = false
+
+                r
+            end
```
doudou: The point of the Ruby `do_something_with_some_config { }` pattern is to
- set the configuration to some known value
- yield
- return the configuration (1) as it was and (2) in all cases (including if an exception is raised)

For (1), you would need to save the value of `@merge_when_identical_agents` and restore it afterwards to make the method really worth it. For (2), you should be guarding the yield in a begin ... ensure block:

```
def merge_task_contexts...
    old_value = @merge_when
    @merge_when = enabled
    yield
ensure
    @merge_when = old_value
end
```

I really don't like the need for this flag ... but let's see another time how we can solve it.

lib/syskit/network_generation/merge_solver.rb
```diff
@@ -220,16 +216,49 @@ def may_merge_task_contexts?(merged_task, task)
...
                 end
 
                 true
             end
 
+            def mergeable_agents?(merged_task, task)
+                unless merge_when_identical_agents?
+                    return true unless task.execution_agent && merged_task.execution_agent
+
+                    return false
```
doudou: This is `return !(task.execution_agent && merged_task.execution_agent)`, which is a lot easier to parse.

lib/syskit/exceptions.rb
```diff
@@ -483,6 +483,39 @@ def pretty_print(pp)
...
+                pp.breakable
+                pp.text "because the same "
+                process_server_name = deployed_task.configured_deployment
+                                                   .process_server_name
+                orogen_model = deployed_task.configured_deployment
+                                            .orogen_model
+                pp.text(
+                    "deployed task #{deployed_task.mapped_task_name} from deployment " \
+                    "#{orogen_model.name} is defined in " \
+                    "#{orogen_model.project.name} on #{process_server_name}"
```
doudou: That a task is defined somewhere else does not impede anything.

The issue is that the same deployment is used multiple times. Since we demand that all deployments are given in the deployment stage, you should deploy the task and the corresponding toplevel tasks (i.e. explain why these tasks are in the plan)

I recommend discussing this with @jhonasiv, he has a very good understanding now of the relationship between tasks and instance requirements :P

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -343,6 +386,36 @@ def self.verify_device_allocation(plan, toplevel_tasks_to_requirements = {})
...
+                    deployed_task = deployed_tasks.select do |d|
+                        d.mapped_task_name == orocos_name
+                    end
+
+                    h[deployed_task.first] = tasks
+                end
+
+                raise ConflictingDeploymentAllocation.new(
+                    deployment_to_task
+                ), "there are deployments used multiple times"
```
doudou: Need to change, cf comment about ConflictingDeploymentAllocation

lib/syskit/network_generation/engine.rb
wvmcastro: Jhonas asked the same thing .. I did this way because it seems that all these parameters are propagated from `#resolve_system_network`, and there I use `Syskit.conf.early_deploy?`. But it shouldn't harm putting it here too

lib/syskit/exceptions.rb
```diff
@@ -447,37 +449,49 @@ def initialize(device, task0, task1, toplevel_tasks_to_requirements = {})
...
+            deployment_to_tasks.each do |deployed_task, tasks|
+                process_server_name = deployed_task.configured_deployment
+                                                   .process_server_name
+                orogen_model = deployed_task.configured_deployment.orogen_model
+                pp.text(
+                    "deployed task '#{deployed_task.mapped_task_name}' from deployment " \
+                    "'#{orogen_model.name}' defined in " \
+                    "'#{orogen_model.project.name}' on '#{process_server_name}' is " \
+                    "assigned to multiple tasks. Here follows one merge failure " \
+                    "(it can have more):"
```
doudou: `it may`

lib/syskit/exceptions.rb
```diff
@@ -447,37 +449,49 @@ def initialize(device, task0, task1, toplevel_tasks_to_requirements = {})
...
+                )
+                print_failed_merge_chain(pp, tasks[0], tasks[1])
+                tasks.each do |t|
+                    defs = find_all_related_syskit_actions(
+                        t, toplevel_tasks_to_requirements
+                    )
+                    print_dependent_definitions(pp, t, defs)
                 end
```
doudou: I think it would make more sense to display the list of tasks first, and then the merge failure:

```
assigned to #{tasks.size} tasks. After the list is a detailed explanation of why the first two tasks are not merged:
```

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -343,6 +386,40 @@ def self.verify_device_allocation(plan, toplevel_tasks_to_requirements = {})
...
+                                     .each_with_object({}) do |(orocos_name, tasks), h|
+                    deployed_tasks = default_deployment_group
+                                     .find_all_suitable_deployments_for(tasks.first)
+
+                    deployed_task = deployed_tasks.select do |d|
+                        d.mapped_task_name == orocos_name
+                    end
+
+                    h[deployed_task.first] = tasks
+                end
```
doudou: The deployment is not guaranteed to be coming from the default deployment group.

But, in any case, the information you're looking for is already in Deployment. Specifically, `execution_agent.model.deployed_orogen_model_by_name`. The type won't be the same, but it IMO has what you need.

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -343,6 +386,40 @@ def self.verify_device_allocation(plan, toplevel_tasks_to_requirements = {})
...
+            def self.verify_all_deployments_are_unique(
+                plan,
+                default_deployment_group,
+                toplevel_tasks_to_requirements
+            )
+                deployment_to_task_map = {}
+                plan.find_local_tasks(Syskit::TaskContext).each do |t|
+                    deployment_to_task_map[t.orocos_name] =
+                        (deployment_to_task_map[t.orocos_name] || []) + [t]
+                end
```
doudou: This is Enumerable#group_by

test/network_generation/test_merge_solver.rb
```diff
@@ -48,12 +48,32 @@
...
+        it "returns false if both tasks have execution agents and " \
+           "merge_when_identical_agents is false" do
+            plan.add(task1 = simple_component_model.new)
+            plan.add(task2 = simple_composition_model.new)
+            [task1, task2].permutation.each do |t1, t2|
+                flexmock(t1).should_receive(:execution_agent).and_return(true)
+                t1.should_receive(:can_merge?).with(t2).and_return(true).once
+            end
+            assert !solver.may_merge_task_contexts?(task1, task2)
+            assert !solver.may_merge_task_contexts?(task2, task1)
```
jhonasiv: `assert !<insert your code here>` should be a `refute`

jhonasiv (approved):
Just this small suggestion, I am fine with the rest of the PR

lib/syskit/exceptions.rb
```diff
@@ -447,38 +449,50 @@ def initialize(device, task0, task1, toplevel_tasks_to_requirements = {})
...
+
+        def pretty_print(pp)
+            deployment_to_tasks.each do |orocos_name, tasks|
+                agent = tasks.first.execution_agent
+                deployment_m = agent.deployed_orogen_model_by_name(orocos_name)
+                pp.text(
+                    "deployed task '#{orocos_name}' from deployment " \
+                    "'#{deployment_m.name}' defined in " \
+                    "'#{deployment_m.project.name}' on '#{agent.process_server_name}' " \
+                    "is assigned to #{tasks.size} tasks. Bellow is the list of " \
```
jhonasiv:
```suggestion
                    "is assigned to #{tasks.size} tasks. Below is the list of " \
```

lib/syskit/exceptions.rb
```diff
@@ -447,38 +449,50 @@ def initialize(device, task0, task1, toplevel_tasks_to_requirements = {})
...
+                    "the dependent non-deployed actions. Right after the list is " \
```
jhonasiv:
```suggestion
                    "the dependent non-deployed actions. Right after the list " \
```

Sorry about the split reviews...I am figuring out how to adapt this message to be used in the context of the error capture feature, as it is task-centered, meaning that I have to individualize the error for each involved task.

lib/syskit/exceptions.rb
```diff
@@ -475,24 +475,24 @@ def initialize(deployment_to_tasks, toplevel_tasks_to_requirements = {})
         end
 
         def pretty_print(pp)
-            deployment_to_tasks.each do |deployed_task, tasks|
-                process_server_name = deployed_task.configured_deployment
-                                                   .process_server_name
-                orogen_model = deployed_task.configured_deployment.orogen_model
+            deployment_to_tasks.each do |orocos_name, tasks|
+                agent = tasks.first.execution_agent
+                deployment_m = agent.deployed_orogen_model_by_name(orocos_name)
```
doudou: You need to extract this information in the constructor. There is no guarantee that the plan will be the way it was once pretty_print is called

doudou (approved):

PR #446, opened 2024-12-16
chore: avoid re-resolving a query in dynamic port bindings if the current selection matches

kapeps:
This is done as an initiative to optimize the dynamic port binding update with an early return.
And fix the broken === for the PortMatcher.

Running test in mk1 robot, it went from 399.9295 seconds to 397.8417

### DIFF
--- a/lib/syskit/dynamic_port_binding.rb
+++ b/lib/syskit/dynamic_port_binding.rb
@@ -99,6 +99,10 @@ module Syskit
         #   the port was updated, and false otherwise. The tuple's second element
         #   is the new resolved port which may be nil if no ports can be found
         def update
+            if @resolved_port && @port_resolver&.current_selection_valid?(@resolved_port)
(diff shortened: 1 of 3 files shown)


lib/syskit/dynamic_port_binding.rb
```diff
@@ -319,6 +327,10 @@ def initialize(port)
                 @port = port
             end
 
+            def current_selection_valid?(port)
```
jhonasiv, line 330: there are two version of current_selection_valid...

kapeps: It depends on the resolver, for the ComponentPortResolver and CompositionChildPortResolver, I used the condition in the update, as for the MatcherPortResolver, I checked whether the matcher finds the provided port

jhonasiv (approved):

doudou (approved):

PR #447, opened 2024-12-17
feat: can_deploy_all method in profile assertions

kapeps:
This pr is for creating assert_can_deploy_all method in profile assertions, that deploy in bulk all actions. This test is more strict than assert_can_deploy, as it will enforce that the profile/interface can deploy all of the actions at the same time. It also is much faster than assert_can_deploy, as it runs all deployments at the same time.

### DIFF
--- a/lib/syskit/test/profile_assertions.rb
+++ b/lib/syskit/test/profile_assertions.rb
@@ -206,8 +206,10 @@ module Syskit
                 action_or_profile = subject_syskit_model,
                 message: "%s is not self contained", exclude: [], **instanciate_options
             )
-                actions = assert_actions(action_or_profile, exclude: exclude)
-
+                actions = validate_assert_actions_argument(
+                    action_or_profile,
+                    exclude: exclude
+                )
                 actions.each do |act|
                     syskit_assert_action_is_self_contained(
                         act, message: message, **instanciate_options
@@ -303,10 +305,14 @@ module Syskit
                 action_or_profile = subject_syskit_model,
                 exclude: [], together_with: []
             )
-                actions = assert_actions(action_or_profile, exclude: exclude)
-
-                together_with = assert_together_with(together_with, exclude: exclude)
-
+                actions = validate_assert_actions_argument(
+                    action_or_profile,
+                    exclude: exclude
+                )
+                together_with = validate_assert_together_with_argument(
+                    together_with,
+                    exclude: exclude
+                )
                 actions.each do |action|
                     assert_can_instanciate_together(action, *together_with)
                 end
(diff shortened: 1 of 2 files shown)


lib/syskit/test/profile_assertions.rb
```diff
@@ -308,22 +303,9 @@ def assert_can_instanciate(
                 action_or_profile = subject_syskit_model,
                 exclude: [], together_with: []
             )
-                actions = validate_actions(action_or_profile, exclude: exclude) do |skip|
-                    flunk "could not validate some non-Syskit actions: #{skip}, " \
-                          "probably because of required arguments. Pass the action to " \
-                          "the 'exclude' option of #{__method__}, and add a separate " \
-                          "assertion test with the arguments added explicitly"
-                end
+                actions = assert_actions(action_or_profile, exclude: exclude)
```
doudou: I think the `assert_actions` and `assert_together_with` names are really not that great...

Maybe `validate_assert_actions_argument` and `validate_assert_together_with_argument` ?

lib/syskit/test/profile_assertions.rb
```diff
@@ -465,6 +434,62 @@ def assert_can_deploy_together(*actions)
...
+            # @param together_with test that all actions in `action_or_profile`
+            #   can be instanciated when all actions in `together_with` are instanciated
+            #   at the same time. This can be used if the former depend on the presence
+            #   of the latter, or if you want to test against conflicts.
+            def assert_can_deploy_all(
+                action_or_profile = subject_syskit_model,
+                exclude: [], together_with: []
+            )
+                actions = assert_actions(action_or_profile, exclude: exclude)
+
```
doudou: Newlines are meant to separate "paragraphs" of code. When you have three lines of text, you don't put newlines between them ...

lib/syskit/test/profile_assertions.rb
```diff
@@ -465,6 +434,62 @@ def assert_can_deploy_together(*actions)
...
+            # {assert_can_deploy_together}
+            def assert_can_deploy_all_together(*actions)
+                syskit_run_deploy_in_bulk(
+                    actions.flatten, compute_policies: true, compute_deployments: true
+                )
+            rescue Minitest::Assertion, StandardError => e
+                raise ProfileAssertionFailed.new("deploy all together", actions, e),
+                      e.message, e.backtrace
+            end
+
```
doudou: assert_can_deploy_together is already doing the job, I don't see the need for this one. Just flatten and splat the `actions` arguments in the assert methods.

lib/syskit/test/profile_assertions.rb
```diff
@@ -567,6 +602,37 @@ def validate_actions(action_or_profile, exclude: [])
...
+            # @api private
+            # Assert together_with actions, flunking in case actions could not be
+            # validated,excluding those included in 'exclude'.
+            #
+            # @param together_with an action interface or profile
+            def assert_together_with(together_with, exclude: [])
+                validate_actions(together_with, exclude: exclude) do |skip|
+                    caller_method = caller[2].split("`").last.split("'").first
```
doudou: Ruby 2.7+ has `caller_locations` which returns objects and removes the need for this.

doudou (approved):

jhonasiv (approved):

PR #448, opened 2024-12-19
feat: Transfer logs from a process server to a central system

DeboraTahara:


lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -29,6 +31,19 @@ def initialize(
                 @max_archive_size = max_archive_size
             end
 
+            # Iterate over all datasets in a Roby log root folder and transfer them
+            # through FTP server
+            #
+            # @param [Params] server_params the FTP server parameters:
```
doudou: use a Struct instead of a hash

`FTPParameters = Struct.new(:host, :port, ..., keyword_init: true)`

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -29,6 +31,19 @@ def initialize(
...
+            # { host, port, certificate, user, password, implicit_ftps, max_upload_rate }
+            def process_root_folder_transfer(server_params)
+                candidates = self.class.find_all_dataset_folders(@root_dir)
+                candidates.each do |child|
+                    basename = child.basename.to_s
+                    process_dataset_transfer(basename, server_params)
```
doudou: `basename` here points to the folder basename, not a file name (`candidates` is a `Array<Pathname>` that points to folders as the name "find_all_dataset_folders" indicates). Beyond needing the full path and not the base name, FTPUpload works at the file level.

However, the whole complexity of this class is that we deal with folders where some files can be copied (the ones that are "finished" and some not (the ones that are being updated). The whole logic of what can be processed or not is currently in `archive_dataset`. You need to refactor it to share that logic here in the transfer case, but transferring instead of archiving.

lib/syskit/runtime/remote/server/ftp_upload.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
```
doudou: What is this folder ?

I would put it in `roby_app/log_transfer` to be honest. And move SpawnServer here (renaming it into just Server)

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -2,6 +2,9 @@
 
 require "archive/tar/minitar"
 require "sys/filesystem"
+require "syskit/process_managers/remote/protocol"
```
doudou: You shouldn't need that one ...

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -2,6 +2,9 @@
...
+require "syskit/roby_app/log_transfer_server/ftp_upload"
+require "net/ftp"
```
doudou: And you should definitely not require this one explicitly. If it is needed, it should be required by `ftp_upload`.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -104,6 +123,46 @@ def process_dataset(child, full:)
...
+
+            # Transfer the given dataset
+            def self.transfer_dataset(
+                dataset_path, server,
+                full:, logger: null_logger
+            )
+                logger.info(
+                    "Transfering dataset #{dataset_path} in #{full ? 'full' : 'partial'} mode"
+                )
+                candidates = each_file_from_path(dataset_path).to_a
```
doudou, line 149: Not needed anymore I believe ...

test/cli/test_log_runtime_archive.rb
```diff
@@ -516,14 +516,73 @@ def should_archive_dataset(dataset, archive_basename, full:)
...
+                            @params[:certfile_path],
+                            interface: @params[:interface], port: @params[:port]
+                        )
+                        server.run
+                    end
+                    thread.join
+                end
+
+                def transfer_dataset(ftp, src_path, tgt_path)
+                    ftp.putbinaryfile(src_path, tgt_path)
```
doudou: I'm pretty sure that's not what you want to do in this test ...

doudou (commented):
From where I sit, I think the general structure is valid. Missing tests, but I'm guessing you are aware of that ...

lib/syskit/cli/log_runtime_archive.rb
DeboraTahara: Why not? We still need to have a list of all files.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -39,10 +39,10 @@ def initialize(
             #
             # @param [Params] server_params the FTP server parameters
             def process_root_folder_transfer(server_params)
-                candidates = self.class.find_all_dataset_folders(@root_dir)
```
eduardacoppo: Just noticed this should be kept. The folder names in the tests should match the pattern. Will fix it.

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
```diff
@@ -0,0 +1,111 @@
...
+                end
+
+                def ensure_dataset_path_exists(ftp, root)
+                    dataset_path = File.dirname(@file.relative_path_from(root))
+
+                    dataset_path.split("/") do |folder|
+                        ftp.chdir(folder)
+                    rescue StandardError => e
+                        ftp.mkdir(folder)
+                        ftp.chdir(folder)
```
jhonasiv: Do not rescue this, its better to look for the folder instead, I guess using ftp.ls

lib/syskit/runtime/server/write_only_disk_file_system.rb
```diff
@@ -1,16 +1,18 @@
...
             # Custom write-only file system that detects collision between files
             class WriteOnlyDiskFileSystem
                 include Ftpd::DiskFileSystem::Base
+                include Ftpd::DiskFileSystem::Mkdir
                 include Ftpd::DiskFileSystem::FileWriting
                 include Ftpd::TranslateExceptions
 
                 def initialize(data_dir)
-                    set_data_dir data_dir
+                    # Ftpd base methods expect data_dir to be a string
```
jhonasiv, line 14: You can raise if data_dir doesnt support #to_s, explaining this expectation

test/cli/test_log_runtime_archive.rb
```diff
@@ -516,14 +517,121 @@ def should_archive_dataset(dataset, archive_basename, full:)
...
+                        make_random_file "test.1.log", root: dataset_a
+                        make_random_file "test.0.log", root: dataset_b
+                        make_random_file "test.1.log", root: dataset_b
+
+                        @process.process_root_folder_transfer(@params)
+
+                        assert(File.exist?(@target_dir / "20220434-2023" / "test.0.log"))
+                        assert(File.exist?(@target_dir / "20220434-2023" / "test.1.log"))
+                        assert(File.exist?(@target_dir / "20220434-2024" / "test.0.log"))
+                        refute(File.exist?(@target_dir / "20220434-2024" / "test.1.log"))
```
jhonasiv, line 572: Why wouldnt `test.1.log` be for dataset_b be finished in this test?

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -128,6 +129,118 @@ def call_archive(root_path, archive_path, low_limit, freed_limit)
...
+                    @target_dir = make_tmppath
+                    @server_params = server_params
+                end
+
+                it "creates an FTP server" do
+                    flexmock(Runtime::Server::SpawnServer)
+                        .new_instances
+                        .should_receive(:initialize).with(
+                            @target_dir, *@server_params.values
+                        ).once
```
jhonasiv: This doesnt test anything, I am not even sure how this passes.IMO the best way to test that the server was created is to send a noop command to it after starting it normally through the CLI.

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -128,6 +129,118 @@ def call_archive(root_path, archive_path, low_limit, freed_limit)
...
+                end
+
+                after do
+                    @server.stop
+                    @server.join
+                end
+
+                it "calls transfer with the specified period" do
+                    mock_files_size([])
+                    mock_available_space(200) # 70 MB
```
jhonasiv: Comment is wrong

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
jhonasiv: Actually, after taking a deeper look into this, using ls would make WriteOnlyFileSystem perform a read...and it would also increase the number of ftp requests per candidate to at least 2. I now think that rescuing the chdir is better.

test/cli/test_log_runtime_archive_main.rb
jhonasiv: Opening the connection with the server is fine too

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
```diff
@@ -0,0 +1,112 @@
...
+                #
+                # @return [LogUploadState::Result]
+                def open_and_transfer(root: nil)
+                    open { |ftp| transfer(ftp, root) }
+                    LogUploadState::Result.new(@file, true, nil)
+                rescue StandardError => e
+                    LogUploadState::Result.new(@file, false, e.message)
+                end
+
+                def ensure_dataset_path_exists(ftp, root)
```
jhonasiv:
```suggestion
                def chdir_to_file_directory(ftp, root)
```

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -17,8 +18,12 @@ class CompressionFailed < RuntimeError; end
         class LogRuntimeArchive
             DEFAULT_MAX_ARCHIVE_SIZE = 10_000_000_000 # 10G
 
+            FTPParameters = Struct.new(:host, :port, :certificate, :user, :password,
+                                       :implicit_ftps, :max_upload_rate,
+                                       keyword_init: true)
+
             def initialize(
-                root_dir, target_dir,
+                root_dir, target_dir: nil,
```
wvmcastro, line 34: Document how `target_dir` impact `LogRuntimeArchive` behaviour. It is obvious now, but it won't be once we forget about it

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
```diff
@@ -0,0 +1,112 @@
...
+
+                # Do transfer the file through the given connection
+                #
+                # @param [Net::FTP] ftp
+                # @param [Pathname] root the archive root folder
+                def transfer(ftp, root)
+                    last = Time.now
+                    opening_mode = root ? "w+" : "r"
+                    File.open(@file, opening_mode) do |file_io|
+                        ensure_dataset_path_exists(ftp, root) if root
```
jhonasiv:
```suggestion
                    ensure_dataset_path_exists(ftp, root) if root
                    File.open(@file) do |file_io|                        
```

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -83,9 +102,10 @@ def ensure_free_space(free_space_low_limit, free_space_delete_until)
 
             def process_dataset(child, full:)
                 use_existing = true
+                basename = child.basename.to_s
```
wvmcastro: if you really want to create a variable, I would name it `filename` instead of `basename`. It is more descriptive in my opinion.

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
```diff
@@ -0,0 +1,112 @@
...
+                    end
+                end
+
+                # Do transfer the file through the given connection
+                #
+                # @param [Net::FTP] ftp
+                # @param [Pathname] root the archive root folder
+                def transfer(ftp, root)
+                    last = Time.now
+                    opening_mode = root ? "w+" : "r"
```
wvmcastro: It is not clear for me why the opening mode changes conditioned by root?

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
```diff
@@ -0,0 +1,112 @@
...
+                rescue StandardError => e
+                    LogUploadState::Result.new(@file, false, e.message)
+                end
+
+                def ensure_dataset_path_exists(ftp, root)
+                    dataset_path = File.dirname(@file.relative_path_from(root))
+
+                    dataset_path.split("/") do |folder|
+                        ftp.chdir(folder)
+                    rescue StandardError => _e
```
wvmcastro: This _rescue_ will catch anything, I think we want to catch just FTP related exceptions, i.e. `FTPPermError` https://ruby-doc.org/stdlib-2.4.0/libdoc/net/ftp/rdoc/Net/FTP.html#method-i-chdir

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
wvmcastro: rt

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
```diff
@@ -0,0 +1,111 @@
...
+                    end
+                end
+
+                # Do transfer the file through the given connection
+                #
+                # @param [Net::FTP] ftp
+                # @param [Pathname] root the archive root folder
+                def transfer(ftp, root)
+                    last = Time.now
+                    ensure_dataset_path_exists(ftp, root) if root
```
wvmcastro: it is not clear when root can be _nil_, in fact I would expect it should never be _nil_

wvmcastro (commented):
Did yo forget https://github.com/rock-core/tools-syskit/pull/448#discussion_r1913135056 ? Pass through all comments so far, also mark the resolved ones as resolved, please

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
eduardacoppo: Already answered this in private. But just for the record:
`root` is `nil` in the usage that already existed. It wasn't used before and this can be easily checked by just reading the code.
Why I added this optional argument: The FTP server only copies the file, it does not create the hierarchy of folders the file is under. And we must create this hierarchy, otherwise all log files from all datasets will be in the same place (and some would possibly be overwritten). So I pass root to know exactly the folder hierarchy the file is under, create these folders and copy the file to the correct location.

wvmcastro: > It wasn't used before

Exactly the point I was trying to make. If it is not used why bother maintaining the behaviour?

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -17,8 +18,19 @@ class CompressionFailed < RuntimeError; end
...
 
+            FTPParameters = Struct.new(:host, :port, :certificate, :user, :password,
+                                       :implicit_ftps, :max_upload_rate,
+                                       keyword_init: true)
+
+            # Initializes the LogRuntimeArchive
+            #
+            # @param [Pathname] root_dir the file's source path
+            # @param [Pathname] target_dir the path to store the file in the archive,
+            #   which is nil in transfer mode, as the target_dir is the FTP server
```
wvmcastro:
```suggestion
            # @param [Pathname] root_dir the logs directory
            # @param [Pathname] target_dir the path to store the file in the archive,
            #   which should be nil in transfer mode, as the logs will be transferred directly to the ftp server @see process_root_folder_transfer
```

wvmcastro (approved):

lib/syskit/roby_app/log_transfer_server/ftp_upload.rb
jhonasiv: > > It wasn't used before
> 
> Exactly the point I was trying to make. If it is not used why bother maintaining the behaviour?

Because the parameter was introduced in this PR. There is still parts of syskit that use the #transfer method so they should continue working.

jhonasiv (approved):

PR #452, opened 2025-01-02
feat: use rock_time_field when logging

kapeps:
Depends on:
 - [ ] https://github.com/rock-core/tools-orogen_syskit_tests/pull/7

Rock's logger already can use a field [to save in the logical time of the stream](https://github.com/rock-core/tools-logger/blob/master/tasks/Logger.cpp#L236). In this PR I modify syskit to pass the "right" field to the logger.

### DIFF
--- a/lib/syskit/network_generation/logger.rb
+++ b/lib/syskit/network_generation/logger.rb
@@ -26,6 +26,20 @@ module Syskit
                 false
             end
 
+            def logical_time_field(type)
+                return unless type < Typelib::CompoundType
+
+                metadata = type.field_metadata
+                type.each_field do |field|
+                    next unless metadata[field].include?("role")
+
+                    role = metadata[field].get("role").first
+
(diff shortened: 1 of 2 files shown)


kapeps:
How should I go on to actually test that the logical_time_field returns the correct field? I am failing to see how to do so using the flexmock @doudou

doudou (commented):
I thiink you should add an integration test for this. Syskit has a `live` test folder that is actually executed with `orogen-test` in `test/live/`. I would create a logger test suite that would use a component from orogen_syskit_tests. That component would pick a static Time value from a property and save it on its output field. Then you can run it and open the log files to check the value.

kapeps:
I meant to ask for another review in the other PR, please disregard this PR if you can read this comment

doudou (commented):
Right ... adding an empty review to get it out of the list.

doudou (approved):

PR #453, opened 2025-01-24
fix: log transfer/transfer_server CLI commands

kapeps:
The Transfer command was passing a hash instead of the FTPParameters, and passing the path instead of the certificate, and the server was not running, only being created.

Tested it by creating an RSA private key:
openssl genrsa -out rsa_private_key 2048

then a certificate:
openssl req -new -x509 -key rsa_private_key -out certificate.pem -subj "/CN=127.0.0.1"

then the required certfile:
cat rsa_private_key certificate.pem > certfile.pem

Then ran the server:
syskit log_runtime_archive transfer_server /home/kappes/test/transfer_dir/ 127.0.0.1 2124 /home/user/test/certfile user pwd true

Then transfered files:
syskit log_runtime_archive transfer 127.0.0.1 2124 /home/user/test/log_dir/ /home/user/test/certificate.pem user pwd true

### DIFF
--- a/test/cli/test_log_runtime_archive_main.rb
+++ b/test/cli/test_log_runtime_archive_main.rb
@@ -225,21 +225,23 @@ module Syskit
                     dataset_tmp_path = make_tmppath
                     root_tmp_path = make_tmppath
 
-                    call_create_server(root_tmp_path, @server_params)
+                    server = call_create_server(root_tmp_path, @server_params)
 
                     make_dataset(dataset_tmp_path, "19981222-1301")
                     make_dataset(dataset_tmp_path, "19981222-1302")
 
-                    call_transfer(dataset_tmp_path)
+                    call_transfer(dataset_tmp_path, server_port: server.port)
                     assert(File.exist?(root_tmp_path / "19981222-1301" / "test.0.log"))
                 end
 
                 # Call 'transfer' function instead of 'watch' to call transfer once
-                def call_transfer(source_dir)
+                def call_transfer(source_dir, server_port: nil)
+                    updated_server_params = @server_params
+                    updated_server_params[:port] = server_port if server_port
                     args = [
                         "transfer",
                         source_dir,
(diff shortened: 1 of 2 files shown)


test/cli/test_log_runtime_archive_main.rb
```diff
@@ -252,7 +291,7 @@ def server_params
                 interface = "127.0.0.1"
                 ca = RobyApp::TmpRootCA.new(interface)
 
-                { host: interface, port: 0,
+                { host: interface, port: 42_429,
```
kapeps: I chose this port randomly, as I need a port different than 0 to test the transfer of files, is this fine, could this generate a bug if this port is in use, how can I avoid it?

wvmcastro: I wouldn't bother before it being a problem, because it is very unlikely. But, I think a possibility is to use `rand(1025...65536)`

wvmcastro (approved):

test/cli/test_log_runtime_archive_main.rb
doudou: It is very definitely NOT possible to use rand(). And it is not very unlikely. The reason the test part of the codebase is peppered with zero ports (when possible) and some scheme of pre-allocation of ports (that is racy but do happen) is because CI would fail regularly because of port collisions.

We're running a lot of tests in parallel that allocate quite a bit of ports, and do so very often. Port collision is bound to happen.

The best is to use port 0 in the allocation of the server, which makes Linux allocate a free port, and then pick up the "real" port from the started server before passing it to the clients.

kapeps: Makes sense, I wasn't able to use port 0, as I chose a port 0 for the server(which was allocated to some port) and then another port 0 in transfer, that would not match the one selected for the server... I simply added an argument in call_transfer that uses the server's port, which at this point has been allocated.

PR #455, opened 2025-01-27
feat: log transfering results

kapeps:
Of working transfer:
image

Of failed transfer:
image

### DIFF
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -189,19 +189,20 @@ module Syskit
             #
             # @result [TransferDatasetResult] the received transfer dataset result
             def self.log_transfer_results(dataset_path, result, logger: null_logger)
-                failed_results = result[:transfer_results].reject do |result|
-                    result.success
-                end
+                failed_results = result[:transfer_results].reject(&:success)
 
                 if failed_results.empty?
                     logger.info(
-                        "Transfering of #{result[:complete] ? "complete" : "incomplete"}"\
-                        " #{dataset_path} finished"
+                        "Transfering of " \
+                        "#{result[:complete] ? 'complete' : 'incomplete'} " \
+                        "#{dataset_path} finished"
                     )
                 else
                     failed_results.each do |failed_result|
-                        failed_message = "with message : " +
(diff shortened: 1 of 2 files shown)


kapeps:
@doudou I know we talked about it being in the CLI, and receiving an array of TransferingDatasetResults, but I found I could do it directly when generating the TransferDatasetResult in transfer_dataset, what do you think?

doudou:
> @doudou I know we talked about it being in the CLI, and receiving an array of TransferingDatasetResults, but I found I could do it directly when generating the TransferDatasetResult in transfer_dataset, what do you think?

Generally speaking, I prefer that library code does not display stuff on the terminal.

What you could do is what most of e.g. syskit-log does: explicitly pass a logger object to the method(s) and use that for display, and create a null logger object that does nothing (which is used by default). This way the library stays silent by default but it's easy to get something output.

kapeps:
> Generally speaking, I prefer that library code does not display stuff on the terminal.
> 
> What you could do is what most of e.g. syskit-log does: explicitly pass a logger object to the method(s) and use that for display, and create a null logger object that does nothing (which is used by default). This way the library stays silent by default but it's easy to get something output.

When you say "explicitly pass" you mean create a flag in the CLI that will create/ignore the logger, or just as it is implemented, where I use the logger from transfer_dataset which is implemented with this default null logger(that is created in the CLI, but offers no choice but to use it)? I am failing to see how my implementation diverges from most of syskit-log.

doudou (approved):

doudou:
> > Generally speaking, I prefer that library code does not display stuff on the terminal.
> > What you could do is what most of e.g. syskit-log does: explicitly pass a logger object to the method(s) and use that for display, and create a null logger object that does nothing (which is used by default). This way the library stays silent by default but it's easy to get something output.
> 
> When you say "explicitly pass" you mean create a flag in the CLI that will create/ignore the logger, or just as it is implemented, where I use the logger from transfer_dataset which is implemented with this default null logger(that is created in the CLI, but offers no choice but to use it)? I am failing to see how my implementation diverges from most of syskit-log.

You're right. This is what I meant ... I replied to the question, but did not look at the PR. Sorry.

PR #456, opened 2025-01-27
feat: ensure_free_space and watch cli commands

kapeps:
Created optional way to overwrite the directory of ensure_free_space, as well as CLI commands that use it for ensuring it in specific directories.

### DIFF
--- a/lib/syskit/cli/log_runtime_archive_main.rb
+++ b/lib/syskit/cli/log_runtime_archive_main.rb
@@ -151,7 +151,7 @@ module Syskit
                 archiver = make_archiver(source_dir)
 
                 source_dir.children.select(&:directory?).sort_by(&:mtime).each do |child|
-                    archiver.ensure_free_space(
+                    break if archiver.ensure_free_space(
                         options[:free_space_low_limit] * 1_000_000,
                         options[:free_space_freed_limit] * 1_000_000,
                         directory: (source_dir / child)
--- a/lib/syskit/cli/log_runtime_archive_main.rb
+++ b/lib/syskit/cli/log_runtime_archive_main.rb
@@ -150,11 +150,13 @@ module Syskit
 
                 archiver = make_archiver(source_dir)
 
-                archiver.ensure_free_space(
-                    options[:free_space_low_limit] * 1_000_000,
-                    options[:free_space_freed_limit] * 1_000_000,
-                    directory: source_dir
-                )
+                source_dir.children.select(&:directory?).sort_by(&:mtime).each do |child|
+                    archiver.ensure_free_space(
+                        options[:free_space_low_limit] * 1_000_000,
+                        options[:free_space_freed_limit] * 1_000_000,
+                        directory: (source_dir / child)
+                    )
+                end
             end
 
             no_commands do # rubocop:disable Metrics/BlockLength
--- a/lib/syskit/cli/log_runtime_archive_main.rb
+++ b/lib/syskit/cli/log_runtime_archive_main.rb
@@ -114,6 +114,49 @@ module Syskit
                               implicit_ftps)
             end
 
+            desc "watch_ensure_free_space", "watches the ensure free space process"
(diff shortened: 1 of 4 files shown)


jhonasiv (approved):

lib/syskit/cli/log_runtime_archive_main.rb
```diff
@@ -114,6 +114,49 @@ def transfer_server( # rubocop:disable Metrics/ParameterLists
...
+                                       deleting files"
+            option :max_size,
+                   type: :numeric, default: 10_000, desc: "max log size in MB"
+            option :free_space_low_limit,
+                   type: :numeric, default: 5_000, desc: "start deleting files if \
+                    available space is below this threshold (threshold in MB)"
+            option :free_space_freed_limit,
+                   type: :numeric, default: 25_000, desc: "stop deleting files if \
+                    available space is above this threshold (threshold in MB)"
+            def ensure_free_space(source_dir)
```
doudou, line 148: For the log transfer, you need to recurse one level inside folders. The `source_dir` should be the root where log directories are created, while in the log archiving we're looking at the archive dir.

Actually, there is also the issue of how we decide which file to delete. In the archive case, we assumed names would be alphabetically sorted in time (they're not, read on). In this case, we deal with arbitrary files.

ensure_free_space should now be using the modification time. Which would also solve the bug on the archive case, where `xxxx.10.tar` would be deleted before `xxxxx.2.tar`

kapeps: Just so we are clear, I also used this modification time when deciding what order I would ensure free space in the source dir, meaning that older directories would be affected first by deletion.

doudou: Right.

lib/syskit/cli/log_runtime_archive_main.rb
```diff
@@ -114,6 +114,51 @@ def transfer_server( # rubocop:disable Metrics/ParameterLists
...
+                source_dir = validate_directory_exists(source_dir)
+
+                archiver = make_archiver(source_dir)
+
+                source_dir.children.select(&:directory?).sort_by(&:mtime).each do |child|
+                    archiver.ensure_free_space(
```
doudou: If you look at a big disk, the command would end up removing very old logs in a folder with a *lot* of folders. It would be a lot better to stop as soon as there's enough free space.

I would recommend to have ensure_free_space return a true/false to indicate whether or not there is enough free space still, and stop if there is.

doudou: Another thing to keep things clean: this loop or the one for data transfer should delete empty folders.

I'd say the data transfer one, actually. The archive loop should as well. Maybe subject for the next PR.

lib/syskit/cli/log_runtime_archive_main.rb
```diff
@@ -114,6 +114,51 @@ def transfer_server( # rubocop:disable Metrics/ParameterLists
                               implicit_ftps)
             end
 
+            desc "watch_ensure_free_space", "watches the ensure free space process"
+            option :period,
+                   type: :numeric, default: 10, desc: "polling period in seconds"
+            option :max_size,
+                   type: :numeric, default: 10_000, desc: "max log size in MB"
```
doudou, line 121: Right. So this is here because the archiver class needs it at construction time. But it's relevant only for archiving.

Now that we have other uses for it, it should be passed as argument to the archive methods instead of given to .new

kapeps: Going to refactor this in a PR on top of this one(removing the max_size from ensure_free_space and transfer related CLI).

kapeps: https://github.com/rock-core/tools-syskit/pull/458

doudou (approved):

PR #459, opened 2025-02-06
feat: init policy defaults to true

eduardacoppo:
The `:init` policy flag requests that the connection sends the last value written to it.
This PR applies the `:init` policy using the `recommend_init` flag

Depends on:
- [ ] https://github.com/rock-core/tools-orogen/pull/23

### DIFF
--- a/lib/syskit/dynamic_port_binding.rb
+++ b/lib/syskit/dynamic_port_binding.rb
@@ -199,10 +199,7 @@ module Syskit
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
-                if port.respond_to?(:init_policy) &&
-                   [true, false].include?(port.init_policy)
-                    policy.merge(init: port.init_policy)
-                end
+                policy.merge(init: policy.fetch(:init, port.model.init_policy))
                 port.reader(**policy)
             end
 
--- a/lib/syskit/dynamic_port_binding.rb
+++ b/lib/syskit/dynamic_port_binding.rb
@@ -199,6 +199,10 @@ module Syskit
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
+                if port.respond_to?(:init_policy) &&
+                   [true, false].include?(port.init_policy)
+                    policy.merge(init: port.init_policy)
+                end
                 port.reader(**policy)
             end
 
--- a/lib/syskit/models/port.rb
+++ b/lib/syskit/models/port.rb
@@ -105,10 +105,6 @@ module Syskit
             # @raise [SelfConnection]
             def connect_to(in_port, policy = {})
                 out_port = to_component_port
-                if out_port.respond_to?(:init_policy) &&
-                    [true, false].include?(out_port.init_policy)
-                    policy = policy.merge(init: out_port.init_policy)
-                end
                 if out_port == self
                     if in_port.respond_to?(:to_component_port)
                         in_port = in_port.to_component_port
--- a/lib/syskit/models/port.rb
+++ b/lib/syskit/models/port.rb
@@ -111,16 +111,7 @@ module Syskit
                 if out_port == self
                     if in_port.respond_to?(:to_component_port)
                         in_port = in_port.to_component_port
-                        if !out_port.output?
-                            raise WrongPortConnectionDirection.new(self, in_port), "cannot connect #{out_port} to #{in_port}: #{out_port} is not an output port"
-                        elsif !in_port.input?
-                            raise WrongPortConnectionDirection.new(self, in_port), "cannot connect #{out_port} to #{in_port}: #{in_port} is not an input port"
-                        elsif out_port.component_model == in_port.component_model
-                            raise SelfConnection.new(out_port, in_port), "cannot connect #{out_port} to #{in_port}: they are both ports of the same component"
-                        elsif out_port.type != in_port.type
-                            raise WrongPortConnectionTypes.new(self, in_port), "cannot connect #{out_port} to #{in_port}: types mismatch"
-                        end
-
+                        validate_connection!(out_port, in_port)
                         component_model.connect_ports(in_port.component_model, [out_port.name, in_port.name] => policy)
                     else
                         Syskit.connect self, in_port, policy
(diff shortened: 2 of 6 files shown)


lib/syskit/models/port.rb
```diff
@@ -105,19 +105,13 @@ def to_component_port
             # @raise [SelfConnection]
             def connect_to(in_port, policy = {})
                 out_port = to_component_port
+                if out_port.respond_to?(:recommend_init)
+                    policy[:init] = out_port.recommend_init
```
doudou: You shouldn't be setting this if `init` was given explicitly by the caller. And you really shouldn't be changing the argument (i.e. should be `policy = policy.merge(init: out_port.recommend_init)`)

test/models/test_port.rb
```diff
@@ -72,6 +72,23 @@
                 out_task_m.out_port.connect_to in_task_m.in_port
             end
         end
+        it "adds 'init: true' as a default policy" do
+            policy = {}
+            flexmock(out_task_m).should_receive(:connect_ports).explicitly.once
+                                .with(in_task_m, %w[out in] => policy.merge(init: true))
```
doudou: `policy` here is empty ... `policy.merge(init: true)` is `{ init: true }`

lib/syskit/models/port.rb
```diff
@@ -105,19 +105,13 @@ def to_component_port
...
-                            raise WrongPortConnectionDirection.new(self, in_port), "cannot connect #{out_port} to #{in_port}: #{out_port} is not an output port"
-                        elsif !in_port.input?
-                            raise WrongPortConnectionDirection.new(self, in_port), "cannot connect #{out_port} to #{in_port}: #{in_port} is not an input port"
-                        elsif out_port.component_model == in_port.component_model
-                            raise SelfConnection.new(out_port, in_port), "cannot connect #{out_port} to #{in_port}: they are both ports of the same component"
-                        elsif out_port.type != in_port.type
-                            raise WrongPortConnectionTypes.new(self, in_port), "cannot connect #{out_port} to #{in_port}: types mismatch"
-                        end
-
+                        validate_connection!(out_port, in_port)
```
doudou: Why the `!` here ?

lib/syskit/models/port.rb
```diff
@@ -105,19 +105,13 @@ def to_component_port
             # @raise [SelfConnection]
             def connect_to(in_port, policy = {})
                 out_port = to_component_port
+                if out_port.respond_to?(:recommend_init)
```
doudou: This guard is very suspicious to me.

doudou (commented):
Good first step.

However, I believe this won't work for ports that are exported on a composition (since they are not using this codepath). I don't think it would work in many cases related to compositions (such as e.g. a composition that has a service as child and gets its real component as an injection in profiles).

The more generic place to resolve this would be where we compute the final connection policy. https://github.com//rock-core/tools-syskit/blob/b674ba903538d229ded38c5980cfd5549535e4eb/lib/syskit/network_generation/dataflow_dynamics.rb#L525

And it will definitely not work for data readers. For data readers, you would need to work here: https://github.com//rock-core/tools-syskit/blob/6b0d92aa0c8bc62a79ce6d58dffb330abdfcceb9/lib/syskit/dynamic_port_binding.rb#L205

lib/syskit/models/port.rb
eduardacoppo: There's a test that tries to do a connection from an input port. Without this guard, the test breaks here.
The test:
image

lib/syskit/models/port.rb
eduardacoppo: This is subjective, but while reading about `!` methods in ruby, I learned that this is a convention in rails for methods that might raise exceptions (example: https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-save-21). Seemed like a practical way to indicate that the method might interrupt the normal flow of code.
But now that you mention it, maybe it doesn't make sense here, since it's not being called by the user. It does draw attention to the method, though, and that was the main idea.

lib/syskit/dynamic_port_binding.rb
```diff
@@ -203,6 +203,10 @@ def initialize(
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
+                if port.respond_to?(:init_policy) &&
```
doudou: Why that check ? It's an output port, it should have the policy.

lib/syskit/dynamic_port_binding.rb
```diff
@@ -203,6 +203,10 @@ def initialize(
...
+                   [true, false].include?(port.init_policy)
+                    policy.merge(init: port.init_policy)
```
doudou: This overrides an explicit `init` flag that may be given in `policy`. The explicit value (from `policy` should take precedence

lib/syskit/dynamic_port_binding.rb
```diff
@@ -203,6 +203,10 @@ def initialize(
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
+                if port.respond_to?(:init_policy) &&
+                   [true, false].include?(port.init_policy)
```
doudou: Do not check this.

The `init_policy` is a default. Use it as a default unconditionally and let the caller override it via the various `policy` arguments.

lib/syskit/models/port.rb
```diff
@@ -105,19 +105,14 @@ def to_component_port
             # @raise [SelfConnection]
             def connect_to(in_port, policy = {})
                 out_port = to_component_port
+                if out_port.respond_to?(:init_policy) &&
+                    [true, false].include?(out_port.init_policy)
+                    policy = policy.merge(init: out_port.init_policy)
+                end
```
doudou: This should not be done here at all. Just do it
- in data reader creation
- in the system network generation

lib/syskit/dynamic_port_binding.rb
eduardacoppo: See https://github.com/rock-core/tools-syskit/pull/459#discussion_r1945282078

I could validate that's an output port to make it more explicit.
`unless port.input?`

lib/syskit/models/port.rb
doudou: It's a ... long discussion in the Ruby circles. The `!` is generally meant to represent "unexpectedly dangerous methods".... Not a very easy criteria to apply. For instance, in containers, it's used to mark self-modifying versions of modification methods (e.g. Hash#merge! vs. Hash#merge)

Regardless of the overall discussion, it's best to always follow the conventions of the codebase you're modifying. Within Syskit, the `validate_*` methods all raise and none has a `!`.

lib/syskit/network_generation/dataflow_dynamics.rb
```diff
@@ -601,17 +601,21 @@ def policy_for(
                         "#{sink_task}:#{sink_port.name}"
                 end
 
+                source_port_m = source_port.model
+                policy ||= {}
```
doudou: You are defining the `policy` local variable here ... It is nil on definition. I.e. these two lines are actually equivalent to

```
policy = { init: source_port_m.init_policy }
```

But then you're merging that in all the policies computed later ... would make more sense to do

```
if ...
    policy = compute_reliable...
elsif
else
end

policy.merge(init: source_port_m.init_policy)
```

doudou (commented):
Right ... so turns out that the syskit dataflow comptuation code itself is binary. Either the policy is empty and it gets computed automatically, or it's not and policy_for gets called.

Let's merge this PR following that behaviour and fix it later.

doudou (approved):

lib/syskit/network_generation/dataflow_dynamics.rb
```diff
@@ -603,23 +603,27 @@ def policy_for(
...
-                    policy
                 else
                     raise UnsupportedConnectionType,
                           "unknown required connection type " \
                           "#{sink_port_m.required_connection_type} " \
                           "on #{sink_port}"
                 end
+
+                source_port_m = source_port.model
+                unless source_port_m.init_policy.nil?
```
jhonasiv:
```suggestion
                unless source_port_m.init_policy?
```

jhonasiv (approved):

PR #460, opened 2025-02-17
chore: warn in case it could not find unmanaged task

kapeps:
This is helpful for debugging purposes.

### DIFF
--- a/lib/syskit/process_managers/unmanaged/process.rb
+++ b/lib/syskit/process_managers/unmanaged/process.rb
@@ -111,22 +111,29 @@ module Syskit
                 # Calls the name service until all of the tasks are resolved. Ignores
                 # whenever a Orocos::NotFound exception is raised.
                 #
+                # @param [Float] warning_period period for warning message in seconds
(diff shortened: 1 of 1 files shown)


lib/syskit/process_managers/unmanaged/process.rb
```diff
@@ -122,6 +122,7 @@ def name_service_get_all_tasks
                         expected_names.delete_if do |name|
                             result[name] = name_service.get(name)
                         rescue Orocos::NotFound
+                            ::Robot.warn "could not find unmanaged task #{name}"
```
doudou: Would be best to rate-limit this message to 1 every 5s, or we will get spammed hard each time the simulation is off.

Could you also add the `sleep 0.1` ? I would suggest doing it only if the first pass did not work (would need to refactor the loop to be clean). This would avoid the current 100% CPU we have when the simulation is not available.

jhonasiv (approved):

doudou (approved):

PR #462, opened 2025-02-17
feat: add flag to set init policy to true

eduardacoppo:
https://github.com/rock-core/tools-syskit/pull/459 shouldn't have been merged. I was still working on a failed test.

### DIFF
--- a/lib/syskit/dynamic_port_binding.rb
+++ b/lib/syskit/dynamic_port_binding.rb
@@ -203,7 +203,7 @@ module Syskit
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
-                policy.merge(init: policy.fetch(:init, port.model.init_policy))
+                @policy.merge(init: @policy.fetch(:init, port.model.init_policy?))
                 port.reader(**policy)
             end
 
--- a/lib/syskit/dynamic_port_binding.rb
+++ b/lib/syskit/dynamic_port_binding.rb
@@ -203,6 +203,10 @@ module Syskit
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
+                if port.respond_to?(:init_policy) &&
+                   [true, false].include?(port.init_policy)
+                    policy.merge(init: port.init_policy)
+                end
                 port.reader(**policy)
             end
 
--- a/lib/syskit/dynamic_port_binding.rb
+++ b/lib/syskit/dynamic_port_binding.rb
@@ -203,10 +203,7 @@ module Syskit
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
-                if port.respond_to?(:init_policy) &&
-                   [true, false].include?(port.init_policy)
-                    policy.merge(init: port.init_policy)
-                end
+                policy.merge(init: policy.fetch(:init, port.model.init_policy))
                 port.reader(**policy)
             end
 
--- a/lib/syskit/dynamic_port_binding.rb
+++ b/lib/syskit/dynamic_port_binding.rb
@@ -203,7 +203,7 @@ module Syskit
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
-                @policy.merge(init: @policy.fetch(:init, port.model.init_policy?))
+                @policy = { init: port.model.init_policy? }.merge(@policy)
                 port.reader(**policy)
             end
 
--- a/lib/syskit/network_generation/dataflow_dynamics.rb
+++ b/lib/syskit/network_generation/dataflow_dynamics.rb
@@ -620,10 +620,7 @@ module Syskit
                 end
 
                 source_port_m = source_port.model
-                unless source_port_m.init_policy.nil?
-                    policy.merge(init: source_port_m.init_policy)
-                end
-                policy
+                policy.merge(init: source_port_m.init_policy?)
             end
 
             def compute_reliable_connection_policy(
--- a/lib/syskit/network_generation/dataflow_dynamics.rb
+++ b/lib/syskit/network_generation/dataflow_dynamics.rb
@@ -601,30 +601,26 @@ module Syskit
                         "#{sink_task}:#{sink_port.name}"
                 end
 
-                source_port_m = source_port.model
-                if [true, false].include?(source_port_m.init_policy)
-                    policy = policy.merge(init: source_port_m.init_policy)
-                end
-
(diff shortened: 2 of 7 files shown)


lib/syskit/dynamic_port_binding.rb
```diff
@@ -203,6 +203,9 @@ def initialize(
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
+                if port.model.init_policy?
+                    policy = policy.merge(init: port.model.init_policy)
+                end
```
doudou: I don't remember how this was ... but

- it should fail with a nomethoderror (`policy` is nil in there), as in `policy.merge` refers to the local variable and not the method from the accessor.
- Unless I am mistaken, it should be model.init_policy?
- it overrides the explicit policy with the default, and it should be the other way around. That is, you should do `{ init: default_value }.merge(policy)` and not the other way.

You can use `@policy` to refer to the policy given to the constructor.

That's a lot of errors for 5 lines of code after a failed merge...

lib/syskit/network_generation/dataflow_dynamics.rb
```diff
@@ -620,10 +620,9 @@ def policy_for(
                 end
 
                 source_port_m = source_port.model
-                unless source_port_m.init_policy.nil?
-                    policy.merge(init: source_port_m.init_policy)
-                end
-                policy
+                return policy unless source_port_m.init_policy?
```
doudou: Once more. Always set the policy. It makes no sense to not set it to false in this case.

lib/syskit/dynamic_port_binding.rb
eduardacoppo: > I don't remember how this was ...

These changes are new, previously I was always setting the policy.

Misunderstood you earlier, policy should keep being set. Got it.

lib/syskit/dynamic_port_binding.rb
```diff
@@ -203,6 +203,7 @@ def initialize(
             # Method called by {Accessor} to create the accessor object from a
             # port
             def create_accessor(port)
+                @policy.merge(init: @policy.fetch(:init, port.model.init_policy?))
```
doudou: - `merge` is returning a new hash. This line does nothing.
- the best way to control priority with `merge` is to decide what hash gets merged into which, essentially: `low_priority.merge(high_priority)`

test/test_dynamic_port_binding.rb
```diff
@@ -382,6 +382,45 @@ module Syskit
...
+                    .and_return({})
+
+                @accessor.create_accessor(@task.out_port)
+            end
+
+            it "expects init: true policy if init_policy(true) is called" do
+                @task.out_port.model.init_policy(true)
+                flexmock(@task.out_port)
+                    .should_receive(:reader)
+                    .and_return({ init: true })
```
doudou: Should be `with` and not `and_return`. Which is why the tests pass even though the implementation is wrong.

test/network_generation/test_dataflow_dynamics.rb
```diff
@@ -380,15 +420,53 @@ module NetworkGeneration
...
+
+                it "merges init policy when sink requires reliable connection" do
+                    @sink_task_m.in_port.needs_reliable_connection
+                    @source_t.out_port.model.init_policy(true)
+
+                    fallback_policy = flexmock
+                    flexmock(@dynamics)
+                        .should_receive(:compute_reliable_connection_policy)
+                        .with(@source_t.out_port, @sink_t.in_port, fallback_policy)
+                        .once.and_return({ init: true })
```
doudou: *You* are returning `init: true` from `compute_reliable_connection_policy`. Not sure what it is meant to test.

jhonasiv:
```suggestion
                        .once
```

test/network_generation/test_dataflow_dynamics.rb
```diff
@@ -380,15 +420,53 @@ module NetworkGeneration
...
+
+                    assert_equal true, policy[:init]
+                end
+
+                it "merges init policy when sink requires 'buffer' connection type" do
+                    @sink_task_m.in_port.needs_buffered_connection
+
+                    flexmock(@source_t.out_port.model)
+                        .should_receive(:init_policy?).explicitly
+                        .and_return(true)
```
jhonasiv, line 461:
```suggestion

```

test/network_generation/test_dataflow_dynamics.rb
```diff
@@ -380,15 +420,53 @@ module NetworkGeneration
...
+
+                    @source_t.out_port.model.init_policy(true)
+                    policy = @dynamics.policy_for(@source_t, "out", "in", @sink_t, nil)
+
+                    assert_equal true, policy[:init]
```
jhonasiv:
```suggestion
                    assert policy[:init]
```

jhonasiv (approved):

test/network_generation/test_dataflow_dynamics.rb
doudou: She still needs to return something.

test/network_generation/test_dataflow_dynamics.rb
doudou: ??? In that case, it makes sense, no ?

test/network_generation/test_dataflow_dynamics.rb
```diff
@@ -380,15 +420,51 @@ module NetworkGeneration
...
+
+                    assert policy[:init]
+                end
+
+                it "merges init policy when sink requires 'buffer' connection type" do
+                    @sink_task_m.in_port.needs_buffered_connection
+
+                    flexmock(@source_t.out_port.model)
+                        .should_receive(:init_policy?).explicitly
+                        .and_return({})
```
doudou: `init_policy?` returns `true` or `false`. The `true` here was correct.

What you're doing with `and_return({})` is put a "true-ish" value.

test/network_generation/test_dataflow_dynamics.rb
jhonasiv: Yeah, I talked with her in person and it did make sense, forgot to update the review in here

test/network_generation/test_dataflow_dynamics.rb
jhonasiv: Also talked in-person, we decided that in here an empty hash works

doudou (approved):

PR #463, opened 2025-02-20
feat: transfer also removes empty log folders

eduardacoppo:
[sc-67798]

Tested it by writing unit tests and following the tutorial [here](https://github.com/rock-core/tools-syskit/pull/457)

### DIFF
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -199,8 +199,7 @@ module Syskit
             end
 
             def self.remove_dataset_folder_if_empty(dataset_path, logger)
-                remaining_files = Dir.children(dataset_path) - ["info.yml"]
-                if remaining_files.empty?
+                if Dir.empty?(dataset_path)
                     begin
                         FileUtils.remove_dir(dataset_path)
                         logger.info("Deleted empty dataset folder: #{dataset_path}")
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -199,21 +199,21 @@ module Syskit
             end
 
             def self.folder_unlocked?(files, path)
+                return false if Roby::Application.log_dir_locked?(path)
+
+                return true if files.empty?
+
                 return false if files.size > 1
 
-                return false if files.first&.end_with?(".lock.tmp")
+                return false if files.first.end_with?(".lock.tmp")
 
-                pp Roby::Application.log_dir_locked?(path)
-                if !Roby::Application.log_dir_locked?(path)
-                    return true if files.empty? || files.first.end_with?(".lock")
-                end
-                false
+                files.first.end_with?(".lock")
             end
 
             def self.remove_dataset_folder_if_empty_and_unlocked(dataset_path, logger)
                 files = Dir.children(dataset_path)
 
-                if self.folder_unlocked?(files, dataset_path)
+                if folder_unlocked?(files, dataset_path)
                     begin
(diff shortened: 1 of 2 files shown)


lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -185,7 +185,24 @@ def self.transfer_dataset(
                 result = TransferDatasetResult.new(
                     complete: complete, transfer_results: transfer_results
                 )
-                log_transfer_results(dataset_path, result, logger: logger)
+                results = log_transfer_results(dataset_path, result, logger: logger)
+                remove_dataset_folder_if_empty(dataset_path, logger)
+                results
```
doudou: I believe that there is a race condition here, especially for process servers: the folder will have only a `info.yml` until the remote system actually generates some other logs.

I actually believe that with the new system, the folder will be really empty in the end (no info.yml)

kapeps: Would the folder start empty as well, I wonder if it is possible for it to be removed before any file(info.yml or logs) are created.
Or is the info.yml created at the same time as the log folder?

kapeps (approved):

lib/syskit/cli/log_runtime_archive.rb
eduardacoppo: The `info.yml` is created at the same time as the log folder. https://github.com/rock-core/tools-syskit/blob/b71da4edbbef33b2ff338567926f0ffbdfb76230/lib/syskit/process_managers/remote/server/server.rb#L456
This is done in the `add_app_metadata` function. https://github.com/rock-core/tools-roby/blob/678e3c59b8255b48c9fae39c97d98df70940d2de/lib/roby/app.rb#L1595

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -185,7 +185,23 @@ def self.transfer_dataset(
...
+            end
+
+            def self.remove_dataset_folder_if_empty(dataset_path, logger)
```
doudou: Reminder: change this logic to delete the folder if it is not locked and the only file present is `.lock`

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -185,7 +185,37 @@ def self.transfer_dataset(
...
+            def self.folder_unlocked?(files, path)
+                return false if files.size > 1
+
+                return false if files.first&.end_with?(".lock.tmp")
+
+                pp Roby::Application.log_dir_locked?(path)
+                if !Roby::Application.log_dir_locked?(path)
+                    return true if files.empty? || files.first.end_with?(".lock")
+                end
+                false
```
jhonasiv:
```suggestion
                return false if files.size != 1
                
                lock_file = files.first
                return false if lock_file.end_with?(".lock.tmp")
                return false unless Roby::Application.log_dir_locked?(path)

                files.empty? || lock_file.end_with?(".lock")
```

eduardacoppo: > `return false if files.size != 1`

This logic is wrong. If folder is empty it should not return false.

> return false if files.first&.end_with?(".lock.tmp")

This is necessary. I have to check whether the file, if it exists, ends with lock.tmp. Also, if there are no files, the code should go on.

> return false unless Roby::Application.log_dir_locked?(path)

Actually, it should return false IF the dir is locked

Other than that, I like the suggestion's readability. Updated the code.

doudou:
The actual implementation should really be three lines plus one test. Will reimplement.

PR #464, opened 2025-02-21
chore: refactor handling of the actual dataflow

doudou:
This PR's objective is to cleanup the ActualDataFlow graph so that it's easier to change. One main change is that the graph is now not directly available anymore, as it was, so that the rest of the runtime code is not directly dependent on the graph structure.


jhonasiv (approved):

lib/syskit/data_flow.rb
```diff
@@ -469,11 +469,9 @@ def all_inputs_connected?(only_static: false)
                     end
 
                     is_connected =
-                        Runtime::ActualDataFlow
-                            .has_edge?(source_task.orocos_task, orocos_task) &&
-                        Runtime::ActualDataFlow
-                            .edge_info(source_task.orocos_task, orocos_task)
-                            .key?([source_port, sink_port])
+                        Runtime::ActualDataFlow.ports_connected?(
+                            source_task.orocos_task, source_port, orocos_task, sink_port
```
wvmcastro, line 473: I can't see from where `orocos_task` is from. Neither in this refactor or in the old code

lib/syskit/runtime/actual_data_flow_graph.rb
```diff
@@ -96,6 +106,127 @@ def static?(task, port)
...
+                @graph.has_edge?(source_task, sink_task)
+            end
+
+            # Returns information about the connections existing between two tasks
+            #
+            # @param [Orocos::TaskContext] source_task
+            # @param [Orocos::TaskContext] sink_task
+            # @return [{[String,String] => Hash}] mapping of (source_port,sink_port) pairs
+            #    to the policy of the established connection
+            def connections_of_tasks(source_task, sink_task)
```
wvmcastro, line 121: If this isn't a heterogeneous graph, I would go with just `def connections(source_task, sink_task)`

lib/syskit/runtime/actual_data_flow_graph.rb
```diff
@@ -96,6 +106,127 @@ def static?(task, port)
...
+            end
+
+            # List connections to the given ports
+            #
+            # @param [Orocos::Task] orocos_task the task whose input port we're inspecting
+            # @param [Array<String>] port_name the name of the input port
+            # @return [{[Orocos::Task,Orocos::Task] => Array<[String,String]>}] matching
+            #   connections, as a mapping of (source_task, sink_task) to the port
+            #   pair (as names)
+            def input_connections_of_ports(orocos_task, port_names)
```
wvmcastro, line 141: `def inputs_for(orocos_task, port_names)`

lib/syskit/runtime/actual_data_flow_graph.rb
```diff
@@ -96,6 +106,127 @@ def static?(task, port)
...
+            end
+
+            # List connections from the given task
+            #
+            # @param [Orocos::Task] orocos_task the task whose output connections are
+            #   expected
+            # @return [{[Orocos::Task,Orocos::Task] => Array<[String,String]>}] matching
+            #   connections, as a mapping of (source_task, sink_task) to the port
+            #   pair (as names)
+            def output_connections_of_task(orocos_task)
```
wvmcastro, line 164: `def outputs_of(orocos_task)`

lib/syskit/runtime/actual_data_flow_graph.rb
```diff
@@ -96,6 +106,127 @@ def static?(task, port)
...
+            #
+            # @param [Orocos::Task] orocos_task the task whose input port we're inspecting
+            # @param [Array<String>] port_name the name of the input port
+            # @return [{[Orocos::Task,Orocos::Task] => Array<[String,String]>}] matching
+            #   connections, as a mapping of (source_task, sink_task) to the port
+            #   pair (as names)
+            def input_connections_of_ports(orocos_task, port_names)
+                @graph.each_in_neighbour(orocos_task)
+                      .with_object({}) do |source_t, result|
+                    mappings = @graph.edge_info(source_t, orocos_task)
```
wvmcastro, line 144: At `static_input_port_connections` you called the return of `@graph_edge_info` connections, I like it better than `mappings` which I think is innocuous

`connections = @graph.edge_info(source_t, orocos_task)`

lib/syskit/runtime/actual_data_flow_graph.rb
doudou: This is the reason of these '_of_task' and 'of_ports'

The input_connections is of_ports, that is is restricted to some ports.

This gives all outputs of a given task. There's actually output_connections_of_ports that restricts to ports. I think not specifying it on all methods would be confusing.

PR #466, opened 2025-02-25
feat: add fine-grained policy merge

eduardacoppo:
Previously here: https://github.com/tidewise/tools-syskit/pull/3
This was on top of another `tools/syskit` PR

### DIFF
--- a/lib/syskit/network_generation/dataflow_dynamics.rb
+++ b/lib/syskit/network_generation/dataflow_dynamics.rb
@@ -553,13 +553,15 @@ module Syskit
                 policy_graph
             end
 
-            def merge_policy(explicit_policy, computed_policy)
-                merged_policy = computed_policy.merge(explicit_policy)
-
-                if merged_policy[:type] == :data
-                    merged_policy.delete(:size)
-                end
-
+            # Merges two connection policies
+            #
+            # @param [Hash] default_policy the policy that has the lowest
+            #   priority
+            # @param [Hash] explicit_policy the policy that has the highest
+            #   priority
+            def merge_policy(default_policy, explicit_policy)
+                merged_policy = default_policy.merge(explicit_policy)
+                merged_policy.delete(:size) if merged_policy[:type] == :data
                 merged_policy
             end
 
@@ -581,10 +583,6 @@ module Syskit
                     policy_graph[[source_task, sink_task]] = computed_policies
                 end
                 policy_graph
-            end
-
-            def policy_default_init_flag(policy, model)
-                policy.merge(init: model.init_policy?)
             end
 
             def policy_compute_data_element(
@@ -645,9 +643,9 @@ module Syskit
                     end
 
                 model = source_task.find_output_port(source_port_name).model
-                computed_policy = policy_default_init_flag(computed_policy, model)
+                computed_policy[:init] = model.init_policy
 
-                merge_policy(explicit_policy, computed_policy)
+                merge_policy(computed_policy, explicit_policy)
             end
 
             def compute_reliable_connection_policy(
--- a/lib/syskit/network_generation/dataflow_dynamics.rb
+++ b/lib/syskit/network_generation/dataflow_dynamics.rb
@@ -553,6 +553,16 @@ module Syskit
                 policy_graph
             end
 
+            def merge_policy(explicit_policy, computed_policy)
+                merged_policy = computed_policy.merge(explicit_policy)
+
+                if merged_policy[:type] == :data
+                    merged_policy.delete(:size)
+                end
+
+                merged_policy
+            end
+
             # @api private
             #
             # Compute the policies for all connections starting from a given task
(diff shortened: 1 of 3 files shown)


lib/syskit/network_generation/dataflow_dynamics.rb
```diff
@@ -553,6 +553,16 @@ def compute_connection_policies
                 policy_graph
             end
 
+            def merge_policy(explicit_policy, computed_policy)
+                merged_policy = computed_policy.merge(explicit_policy)
+
+                if merged_policy[:type] == :data
```
jhonasiv, line 559: how could this happen? due to a bad explicit policy?

jhonasiv (approved):

doudou (commented):
This is a good first step, but there's one issue.

When using `needs_reliable_connection`, we have to add period information to the syskit models. Until now, setting the policy explicitly would be enough to avoid having to do this (because of time, or because Syskit limitations would make it hard to do). This PR will break these cases.

My suggestion is to split policy_for into:
- policy_compute_data_element(policy, fallback_policy)
- policy_default_init_flag

Always call the latter, but call the former only if the explicit policy does not have `:type` set already. Create corresponding unit tests.

lib/syskit/network_generation/dataflow_dynamics.rb
```diff
@@ -618,9 +627,22 @@ def policy_for(
...
                           "on #{sink_port}"
                 end
+                policy
+            end
+
+            def policy_for(
+                source_task, source_port_name, sink_port_name, sink_task, fallback_policy
+            )
+                policy = {}
+                unless policy[:type]
```
doudou: `policy` is the empty hash here. This will always be nil.

The `if` should obviously be checking the policy that is coming from the policy graph.

I would actually pass `policy` from the loop in `compute_policies_from` to the method, and add the merge_policy call at the end of this method instead of doing it in said loop.

test/network_generation/test_dataflow_dynamics.rb
```diff
@@ -360,6 +363,151 @@ module NetworkGeneration
                     add_agents(tasks[0, 2])
                     flexmock(@dynamics).should_receive(:propagate).with(tasks[0, 2])
                 end
+
+                it "handles the case where the explicit policy sets the type to :data" do
```
doudou, line 369: And what about the type being `:buffer`. We should make sure it's not overriding the size.

test/network_generation/test_dataflow_dynamics.rb
```diff
@@ -360,6 +363,151 @@ module NetworkGeneration
...
+
+                    add_agents(tasks = [task0, task1])
+                    flexmock(@dynamics).should_receive(:propagate).with(tasks)
+
+                    task0.out_port.connect_to task1.in_port, type: :data
+
+                    flexmock(@dynamics)
+                        .should_receive(:policy_for)
+                        .with(task0, "out", "in", task1, nil)
+                        .and_return(type: :buffer, size: 10, init: true)
```
doudou: This is *massively* cheating.

What we want to check is that `policy_compute_data_element` is not called when :type is set. That's the really important bit.

lib/syskit/network_generation/dataflow_dynamics.rb
```diff
@@ -561,25 +571,24 @@ def compute_policies_from(connection_graph, source_task, policy_graph = {})
...
-                                    policy_for(source_task, *port_pair, sink_task,
-                                               fallback_policy)
-                            else
-                                h[port_pair] = policy
-                            end
+                            explicit_policy = policy.dup
+                            fallback_policy = explicit_policy.delete(:fallback_policy)
+                            h[port_pair] = policy_for(
+                                source_task, *port_pair, sink_task,
+                                fallback_policy, explicit_policy: explicit_policy
```
eduardacoppo: `explicit_policy` is a keyword argument because some external tests call `policy_for`, and having it as a positional argument would require changes to those tests as well.

lib/syskit/network_generation/dataflow_dynamics.rb
```diff
@@ -618,9 +627,27 @@ def policy_for(
...
                 end
+                policy
+            end
+
+            def policy_for(
+                source_task, source_port_name, sink_port_name, sink_task,
+                fallback_policy, explicit_policy: nil
+            )
+                computed_policy = {}
+                explicit_policy = {} unless explicit_policy
```
jhonasiv:
```suggestion
                fallback_policy, explicit_policy: {}
            )
                computed_policy = {}
```

You mentioned making this a kw arg due to not wanting to fix the policy_for call alls over the code, but you can make this a positional argument with a default value as well, that would work just like the kw arg worked.

doudou: This. I think that's the last change.

doudou:
@jhonasiv updated, please have a look

jhonasiv (approved):

doudou (commented):
Adding empty review to get it out of my queue (since I'm the one working on it now)

PR #468, opened 2025-02-25
chore: Use logger instead of puts

DeboraTahara:


jhonasiv (approved):

doudou (commented):
You should also make $stdout "sync",

`$stdout.sync = true`

The better implementation would be to have a create_logger method that does both (must be within the no_commands block below) and use it everywhere.

doudou (approved):

PR #469, opened 2025-02-27
feat: add log directory locking and unlocking

eduardacoppo:
[sc-69598]

Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/317

### DIFF
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -46,10 +46,10 @@ module Syskit
             # @return [Array<TransferDatasetResult>]
             def process_root_folder_transfer(server_params)
                 candidates = self.class.find_all_dataset_folders(@root_dir)
-                running = candidates.last
                 candidates.map do |child|
                     process_dataset_transfer(
-                        child, server_params, @root_dir, full: child != running
+                        child, server_params, @root_dir,
+                        full: !Roby::Application.log_dir_locked?(child.basename)
                     )
                 end
             end
@@ -65,10 +65,10 @@ module Syskit
             # @param [Integer] max_archive_size the max size of the archive
             def process_root_folder(max_archive_size: DEFAULT_MAX_ARCHIVE_SIZE)
                 candidates = self.class.find_all_dataset_folders(@root_dir)
-                running = candidates.last
                 candidates.each do |child|
                     process_dataset(
-                        child, max_archive_size: max_archive_size, full: child != running
+                        child, max_archive_size: max_archive_size,
+                               full: !Roby::Application.log_dir_locked?(child.basename)
                     )
                 end
             end
--- a/lib/syskit/process_managers/remote/server/server.rb
+++ b/lib/syskit/process_managers/remote/server/server.rb
@@ -414,23 +414,23 @@ module Syskit
                         socket.write(Marshal.dump(message))
                     end
 
-                    def handle_new_folder_request(metadata = {})
-                        app.unlock_log_dir
+                    def setup_app_name_from_log_dir_metadata(metadata)
+                        return unless (app_name = metadata.dig("parent", "app_name"))
 
-                        return unless (parent_info = metadata["parent"])
+                        app.app_name = app_name
+                    end
 
-                        if (app_name = parent_info["app_name"])
-                            app.app_name = app_name
-                        end
-                        if (robot_name = parent_info["robot_name"])
-                            app.robot(
-                                robot_name, parent_info["robot_type"] || robot_name
-                            )
-                        end
+                    def setup_app_robot_from_log_dir_metadata(metadata)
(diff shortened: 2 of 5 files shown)


lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -157,6 +158,11 @@ def failures
                 end
             end
 
+            # Check if a log folder is locked (indicating an active process)
+            def log_folder_locked?(folder_path)
+                File.exist?(File.join(folder_path, ".lock"))
```
doudou: The existence of the file is not enough. As you *yourself* have implemented in the Roby pull request.

lib/syskit/process_managers/remote/server/server.rb
```diff
@@ -441,20 +447,66 @@ def handle_command(socket) # :nodoc:
...
+                        return unless (robot_name = parent_info["robot_name"])
+
+                        app.robot(robot_name, parent_info["robot_type"] || robot_name)
+                    end
+
+                    def create_log_dir(time_tag, metadata = {})
+                        handle_new_folder_request(metadata)
 
                         app.add_app_metadata(metadata)
                         app.find_and_create_log_dir(time_tag)
```
doudou, line 436: Since this code already uses Application ... how about using the locking code from Applicatio n?

lib/syskit/process_managers/remote/server/server.rb
```diff
@@ -205,6 +209,7 @@ def listen
                                 unless handle_command(socket)
                                     debug "#{socket} closed or errored"
                                     socket.close
+                                    unlock_log_dir
```
doudou: Create a `close_client(socket)` method that does both and use it here and in quit_and_join

lib/syskit/cli/log_runtime_archive.rb
doudou: How is that resolved ?

You are taking a lock. The **existence** of the lock file is not enough. You have to verify that the lock is taken on the file.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -46,10 +46,10 @@ def initialize(
             # @return [Array<TransferDatasetResult>]
             def process_root_folder_transfer(server_params)
                 candidates = self.class.find_all_dataset_folders(@root_dir)
-                running = candidates.last
+                locked = candidates.select { |child| log_folder_locked?(child) }
                 candidates.map do |child|
                     process_dataset_transfer(
-                        child, server_params, @root_dir, full: child != running
+                        child, server_params, @root_dir, full: !locked.include?(child)
                     )
```
doudou: This is

```
full: !log_folder_locked?(child)
```

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -65,10 +65,11 @@ def process_root_folder_transfer(server_params)
             # @param [Integer] max_archive_size the max size of the archive
             def process_root_folder(max_archive_size: DEFAULT_MAX_ARCHIVE_SIZE)
                 candidates = self.class.find_all_dataset_folders(@root_dir)
-                running = candidates.last
```
doudou, line 68: Same

lib/syskit/cli/log_runtime_archive.rb
doudou: Plus, I just explained in great length that a folder that does not have the file should be considered locked.

test/cli/test_log_runtime_archive.rb
```diff
@@ -373,6 +373,7 @@ module CLI
                     dataset0 = make_valid_folder("20220434-2023")
```
doudou: Right. And you will need to update all these tests *and* create tests for `log_folder_locked?` specific for all three possible states (no .lock, .lock with flock, .lock without flock)

lib/syskit/cli/log_runtime_archive.rb
eduardacoppo: Was using the list of locked candidates for debugging. Forgot to simplify the logic afterwards.

lib/syskit/cli/log_runtime_archive.rb
eduardacoppo: > Plus, I just explained in great length that a folder that does not have the file should be considered locked.

The explanation was clear to me. Previously, I had lock and unlock functions here too. The `log_folder_locked?` function went unnoticed when I removed the others.

jhonasiv (approved):

test/cli/test_log_runtime_archive.rb
eduardacoppo: Do I need to have these tests here as well? I wrote them in https://github.com/rock-core/tools-roby/pull/317 and you've kept them in the superseded PR (https://github.com/rock-core/tools-roby/pull/323/)

PR #476, opened 2025-03-12
fix: telemetry UI behaviour on real world tests

doudou:
Fix behaviour of the `telemetry ui` (in its current state).
- fix crash when selecting jobs
- fix discovery behaviour (now prefers resolving tasks of the selected job)
- fix thread safety

This PR does not fix a problem with the display of the task states (it may stay on INITIALIZING while ports can be read already)

### DIFF
--- a/lib/syskit/telemetry/ui/runtime_state.rb
+++ b/lib/syskit/telemetry/ui/runtime_state.rb
@@ -145,6 +145,7 @@ module Syskit
 
                     @current_job = nil
                     @current_job_tasks = []
+                    @current_job_task_names = []
                     @current_tasks = []
 
                     syskit.on_ui_event do |event_name, *args|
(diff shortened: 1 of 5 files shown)


lib/syskit/telemetry/ui/runtime_state.rb
```diff
@@ -669,13 +672,25 @@ def report_app_error(error)
...
                         end
                     end
-                    @name_service.async_update_tasks(all_deployed_tasks)
+
+                    if @current_job
+                        # If we are focussing, keep the tasks that have already been
+                        # discovered, but resolve the ones that interest the user first
+                        focussed_names =
+                            (@name_service.names + @current_job_task_names).to_set
+                        puts "FOCUS #{focussed_names}"
```
jhonasiv: did you intend to leave this puts call?

doudou: Nope

jhonasiv (approved):

PR #467, opened 2025-02-25
fix: rename files during the transfer, adding the .partial extension

DeboraTahara:

PR #477, opened 2025-03-12
feat: rename yml file before transferring (on top of #467)

eduardacoppo:
https://app.shortcut.com/tidewise/story/69064/implement-way-to-change-the-name-of-info-yml-on-process-servers

On top of #467

### DIFF
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -44,12 +44,13 @@ module Syskit
             #
             # @param [Params] server_params the FTP server parameters
             # @return [Array<TransferDatasetResult>]
-            def process_root_folder_transfer(server_params)
+            def process_root_folder_transfer(server_params, info_name: "info")
                 candidates = self.class.find_all_dataset_folders(@root_dir)
                 candidates.map do |child|
                     process_dataset_transfer(
                         child, server_params, @root_dir,
-                        full: !Roby::Application.log_dir_locked?(child.basename)
+                        full: !Roby::Application.log_dir_locked?(child.basename),
+                        info_name: info_name
                     )
                 end
             end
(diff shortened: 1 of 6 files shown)


doudou (commented):
Same comment. Renaming should be done during transfer by log_runtime_archive

jhonasiv (approved):

jhonasiv:
Ah, I approved by accident, but you can consider it an approval after you satisty @doudou comment

doudou:
> Ah, I approved by accident, but you can consider it an approval after you satisty @doudou comment

I essentially said "this pull request should be closed" ...

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -196,6 +198,17 @@ def self.transfer_dataset(
                 log_transfer_results(dataset_path, result, logger: logger)
             end
 
+            # Renames the info.yml file
+            def self.rename_info_file(dataset_path)
+                info_file = dataset_path / "info.yml"
+                return unless info_file.file?
+
+                new_info_file =
+                    dataset_path / "info_#{Time.now.strftime('%Y%m%d-%H%M')}.yml"
```
jhonasiv: Is this from a talk you had in person with someone? Apart from the source of the info not being obvious in the file name (which I think is not that big a deal when the info is mostly used by other tools), I think you still run the risk of conflict if you stop at the "minutes" in the name. You should at least include the seconds in the name (or ms if you are feeling zealous)

doudou: I have no idea why she did this ...

wvmcastro (approved):
In a redundant system we must remember to use names with ids i.e. motion-box-1, ..., motion-box-n

doudou:
> In a redundant system we must remember to use names with ids i.e. motion-box-1, ..., motion-box-n

Good point. Documenting this now (as well as integrating in Ansible).

PR #478, opened 2025-03-12
feat: fails transfer if free space below thresh

eduardacoppo:
https://app.shortcut.com/tidewise/story/64361/stop-log-transfer-is-the-free-space-is-some-threshold-on-the-server

### DIFF
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -150,9 +150,11 @@ module Syskit
                 end
             end
 
-            def process_dataset_transfer(child, server, root, full:, min_required_space:)
+            def process_dataset_transfer(
+                child, server_params, root, full:, min_required_space:
+            )
                 self.class.transfer_dataset(
-                    child, server, root,
+                    child, server_params, root,
                     full: full, min_required_space: min_required_space, logger: @logger
                 )
             end
@@ -171,7 +173,7 @@ module Syskit
 
             # Transfer the given dataset
             def self.transfer_dataset( # rubocop:disable Metrics/ParameterLists
-                dataset_path, server, root,
+                dataset_path, server_params, root,
                 full:, min_required_space:, logger: null_logger
             )
                 logger.info(
@@ -188,7 +190,8 @@ module Syskit
                     end
 
                 transfer_results = candidates.map do |child_path|
-                    result = transfer_file(child_path, server, root, min_required_space)
+                    result = transfer_file(child_path, server_params, root,
+                                           min_required_space)
                     child_path.unlink if result.success?
 
                     result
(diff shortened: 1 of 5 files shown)


lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -18,6 +18,8 @@ class CompressionFailed < RuntimeError; end
         class LogRuntimeArchive
             DEFAULT_MAX_ARCHIVE_SIZE = 10_000_000_000 # 10G
 
+            FREE_SPACE_THRESHOLD_FOR_TRANSFER = 10_000_000_000 # 10G
```
kapeps: I believe this should be an argument of the transfer CLI, where we would be able to set this value when caling it from the command

eduardacoppo: I agree. Will make this an option.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -139,9 +142,10 @@ def process_dataset(child, full:, max_archive_size: DEFAULT_MAX_ARCHIVE_SIZE)
                 end
             end
 
-            def process_dataset_transfer(child, server, root, full:)
+            def process_dataset_transfer(child, server, root, full:, thresh:)
```
kapeps: I believe thresh is a bit undescriptive, I would call it min_required_space(and standarize this variable in the cli, instead of having multiple names for it)

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -43,13 +43,16 @@ def initialize(
             # through FTP server
             #
             # @param [Params] server_params the FTP server parameters
+            # #param [Integer] min_free_space the minimum required space
+            #   in the root_dir to perform transfer
```
kapeps: Is it in bytes? I would use the same unit of space as the other commands i.e. in MB to avoid confusion

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -222,7 +228,16 @@ def self.log_transfer_results(dataset_path, result, logger: null_logger)
             # Transfer a file to the central log server via FTP
             #
             # @return [LogUploadState:Result]
-            def self.transfer_file(file, server, root)
+            def self.transfer_file(file, server, root, target_dir, min_required_space)
+                free_space = Sys::Filesystem.stat(target_dir).bytes_available
```
eduardacoppo: Passing the `target_dir` arg along all these functions is certainly not the best way to do it. Just wanted to leave some fix.

doudou: This is still testing the available bytes on the client (the beaglebone) and not the server (the control computer)

doudou:
This is just wrong. Will open a new one.

PR #479, opened 2025-03-13
fix: use implicit ftps as a boolean option instead of receiving it as a string

kapeps:
1. It was supposed to receive a bool, not a string of true or false.

Tested with real logs:
image

2. There was missing documentation for the cli commands.
image

### DIFF
--- a/lib/syskit/cli/log_runtime_archive_main.rb
+++ b/lib/syskit/cli/log_runtime_archive_main.rb
@@ -72,7 +72,7 @@ module Syskit
             option :max_upload_rate_mbps,
                    type: :numeric, default: 1_000, desc: "max upload rate in Mbps"
             option :implicit_ftps,
-                   type: :boolean, default: true,
+                   type: :boolean, default: false,
                    desc: "use implicit connection method for ftps " \
                          "(disable for 2.5 clients on 2.7 servers)"
             def watch_transfer( # rubocop:disable Metrics/ParameterLists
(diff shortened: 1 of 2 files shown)


kapeps:
@doudou I remember you had further comments regarding the transfer/transfer_server CLI arguments, but I dont remember if it was solely regarding the implicit_ftps, or there was more to it.

doudou (approved):

lib/syskit/cli/log_runtime_archive_main.rb
```diff
@@ -63,19 +65,22 @@ def archive(root_dir, target_dir)
...
-            desc "watch_transfer", "watches a dataset root folder \
-                                    and periodically performs transfer"
+            desc "watch_transfer SOURCE_DIR HOST PORT CERTIFICATE_PATH USER PASSWORD",
+                 "watches a dataset root folder and periodically performs transfer"
             option :period,
                    type: :numeric, default: 600, desc: "polling period in seconds"
             option :max_upload_rate_mbps,
                    type: :numeric, default: 1_000, desc: "max upload rate in Mbps"
+            option :implicit_ftps,
+                   type: :boolean, default: true,
```
doudou: Actually, `false` would be a better default. `--implicit-ftps` is needed because of this 2.5/2.7 mixed situation, but won't be long term.

test/cli/test_log_runtime_archive_main.rb
```diff
@@ -238,11 +242,14 @@ def rate_mbps_to_bps(rate_mbps)
...
                     updated_server_params = @server_params
+                    implicit_ftps = updated_server_params[:implicit_ftps]
+                    updated_server_params.delete(:implicit_ftps)
                     updated_server_params[:port] = server_port if server_port
                     args = [
                         "transfer",
                         source_dir,
-                        *updated_server_params.values
+                        *updated_server_params.values,
+                        "--implicit_ftps", implicit_ftps
```
jhonasiv: would this work? I expect boolean CLI arguments to be added by the flag being there or not, does it parse explicitly called `--bla true` args correctly?

kapeps: I tested it in CLI/unit test and it worked as expected, but you're not the first one asking this, I am going to change to the standard way of boolean CLI arguments.

jhonasiv (approved):

PR #483, opened 2025-03-24
fix: race condition in parallel disconnection

doudou:
Ruby's Hash is *not* thread safe when #hash has been redefined. When the Hash code calls the hash method, the Ruby scheduler may schedule another thread that might create some memory corruption if said thread also modifies the same hash. 

Turns out that Concurrent::Map does not work around this.

This is the case in the disconnection codepath, since Orocos::TaskContext#hash has been modified so that two objects that point to the same remote task look the same as hash keys. So far, this bug only has had an effect in CI, showing up as memory corruption and either a segmentation fault or what looks like type confusion ("an integer does not have method remove_disconnection_from`)

The commit modifies the connection codepath, even though it was not affected (it uses syskit tasks as keys and not orocos tasks, and the syskit tasks do not have their #hash method redefined). This is for symmetry - and to make sure the same problem does not appear along some changes to these codepaths.

### DIFF
--- a/lib/syskit/runtime/connection_management.rb
+++ b/lib/syskit/runtime/connection_management.rb
@@ -231,32 +231,49 @@ module Syskit
             end
 
             def perform_disconnections(disconnections)
-                success = Concurrent::Array.new
-                failure = Concurrent::Array.new
-                port_cache = Concurrent::Map.new
-                promises = disconnections.map do |syskit_from_task, from_task, from_port, syskit_to_task, to_task, to_port|
-                    execution_engine = plan.execution_engine
-                    execution_engine.promise(description: "disconnect #{from_task.name}##{from_port} -> #{to_task.name}##{to_port}") do
-                        begin
-                            from_orocos_port =
-                                (port_cache[[from_task, from_port]] ||= from_task.raw_port(from_port))
-                            to_orocos_port   =
-                                (port_cache[[to_task, to_port]] ||= to_task.raw_port(to_port))
-                            unless from_orocos_port.disconnect_from(to_orocos_port)
-                                warn "while disconnecting #{from_task}:#{from_port} => #{to_task}:#{to_port} returned false"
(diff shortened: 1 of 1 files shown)


lib/syskit/runtime/connection_management.rb
```diff
@@ -231,43 +231,79 @@ def pre_disconnect(removed)
...
+
                     execution_engine = plan.execution_engine
-                    execution_engine.promise(description: "disconnect #{from_task.name}##{from_port} -> #{to_task.name}##{to_port}") do
+                    promise = execution_engine.promise(
+                        description: "disconnect #{from_task.name}##{from_port} -> " \
+                                     "#{to_task.name}##{to_port}"
+                    ) do
                         begin
-                            from_orocos_port =
-                                (port_cache[[from_task, from_port]] ||= from_task.raw_port(from_port))
```
wvmcastro, line 242: Just to be clear that I got it. The problem was that between `port_cache[[from_task, from_port]]` call and completion the `port_cache` object could be modified?

wvmcastro (approved):

jhonasiv (approved):

lib/syskit/runtime/connection_management.rb
doudou: The problem is that inside the `#[]` and `#[]=` calls, that are implemented in C and not in Ruby, the Ruby runtime calls the `#hash` method. That method, in the case of Orocos::TaskContext, is implemented in Ruby.

When `#hash` gets called, Ruby's VM can do a context switch. This created inconsistencies inside the C part of the hash structure and crashes.

A lot of the "thread safety" in MRI (A.k.a. CRuby, a.k.a. the canonical implementation of Ruby) is due to the VM scheduler never doing context switching within the C code. The assumption here - from CRuby implementers and it seems from concurrent-ruby ones - was that it was the case, but that assumption was wrong.

PR #487, opened 2025-04-07
feat: reject log transfers if free space is below threshold (on top of #477)

doudou:
On top of #477

### DIFF
--- a/lib/syskit/runtime/server/write_only_disk_file_system.rb
+++ b/lib/syskit/runtime/server/write_only_disk_file_system.rb
@@ -44,18 +44,18 @@ module Syskit
                     error "Already exists", 550 if File.exist?(final_path)
 
                     # Code copied from Ftpd::DiskFileSystem::FileWriting
-                    verify_free_space
                     File.open(full_path, "wb") do |file|
                         while (line = stream.read)
+                            verify_free_space(line.size)
+
                             file.write line
-                            verify_free_space
                         end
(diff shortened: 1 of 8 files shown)


lib/syskit/runtime/server/write_only_disk_file_system.rb
```diff
@@ -33,15 +36,34 @@ def initialize(data_dir)
...
                     )
                     error "Already exists", 550 if File.exist?(final_path)
 
-                    write_file ftp_path, stream, "wb"
+                    # Code copied from Ftpd::DiskFileSystem::FileWriting
+                    verify_free_space
+                    File.open(full_path, "wb") do |file|
+                        while (line = stream.read)
+                            file.write line
+                            verify_free_space
```
wvmcastro:
```suggestion
                    File.open(full_path, "wb") do |file|
                        while (line = stream.read)
                            verify_free_space_for line.size
                            file.write line
```

and test  `it "does not zero free space if chunk is greater than current free space"`

doudou: It's very academic (I think we should have thresholds of 10GB or more), but I like the idea nonetheless.

We will need to do some accounting of free space (check everything that takes up space and allocate space as needed).

wvmcastro (approved):

jhonasiv (approved):

PR #489, opened 2025-04-07
fix: only emit failed for non-resolved instance requirement tasks

jhonasiv:
Any exception during network planning would kill every single one of the running tasks when the capture errors feature flag is disabled.


doudou (approved):

wvmcastro (approved):

jhonasiv:
Oops, closed by mistake

PR #490, opened 2025-04-07
fix: emit resolution success only once

jhonasiv:
Emitting it more than once makes no difference in the network generation and it being constantly emitted is quite spammy.


doudou:
Discussed with Jhonas and we won't apply this.

PR #492, opened 2025-04-08
fix: validate resolution error processing and garbage collect after cleanup

jhonasiv:
Not validating the resulting plan after cleanup was an obvious issue, it made bugs in the plan afterwards very hard to find. 

One of these bugs was the lack of garbage collection of unneeded tasks after we cleaned up the resolution errors. That meant that we had useless tasks THAT STILL CONTAINED ERRORS in the plan, which would constantly trigger the errors in future deployments. 

Another bug is that there was no update to the required instances between the #compute_deployed_network and #apply_deployed_network_to_plan calls, meaning that wrong required instances would be used to process the network failures (especially when the early_deploy was false)

### DIFF
--- a/lib/syskit/network_generation/engine.rb
+++ b/lib/syskit/network_generation/engine.rb
@@ -112,9 +112,6 @@ module Syskit
                     deployer.deploy(
                         error_handler: error_handler, validate: validate_deployed_network
                     )
-                    required_instances = required_instances.transform_values do |task|
-                        merge_solver.replacement_for(task)
-                    end
                     resolution_errors = error_handler.process_failures(
                         required_instances, cleanup_failed_tasks: true
                     )
@@ -141,7 +138,7 @@ module Syskit
                 @deployment_tasks = work_plan.find_local_tasks(Deployment).to_set
                 @deployed_tasks = work_plan.find_local_tasks(Component).to_set
 
-                [required_instances, resolution_errors]
+                resolution_errors
             end
 
             # Apply the deployed network created with
@@ -761,7 +758,8 @@ module Syskit
                     system_network_generator.toplevel_tasks_to_requirements
 
                 resolution_errors = error_handler.process_failures(
-                    required_instances, cleanup_failed_tasks: cleanup_resolution_errors
+                    required_instances,
+                    cleanup_failed_tasks: cleanup_resolution_errors
                 )
                 if cleanup_resolution_errors
                     # Sanity check that the plan was properly cleaned up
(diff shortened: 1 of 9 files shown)


doudou (approved):

wvmcastro (approved):

lib/syskit/network_generation/resolution_error_handler.rb
```diff
@@ -211,6 +211,7 @@ def cleanup_resolution_errors(
                         # Remove tasks that are not useful anymore
                         @plan.remove_task(obj)
                     end
+                @resolution_failures.clear
```
wvmcastro, line 203: At engine we do not usually reuse other objects states (net gen and net deployer). So, I would just create another error handling object when I needed it

lib/syskit/network_generation/engine.rb
```diff
@@ -752,6 +752,9 @@ def compute_system_network(
                         early_deploy && validate_deployed_network
                 )
                 required_instances = Hash[requirement_tasks.zip(toplevel_tasks)]
+                # Take toplevel tasks to requirements before cleanup
+                toplevel_tasks_to_requirements =
+                    system_network_generator.toplevel_tasks_to_requirements
```
wvmcastro, line 758: I didn't quite get this change. Can you expand on that?  Because I don't see how failure processing can change it

lib/syskit/network_generation/resolution_error_handler.rb
jhonasiv: I did do that in the beginning, but talking with @doudou we decided the right call would be to ensure that the error handler cant reprocess the already processed resolution failures after a cleanup, as they were in fact cleaned up. 

Creating the object again would be a little cumbersome, I proposed creating a method, i.e. `#clear` that just returned a new error handler object maintaining the plan and merge solver, but still, it was fragile as the Engine should use it in the right place (or risk messing up the failure processing)

lib/syskit/network_generation/engine.rb
jhonasiv: We change the plan in cleanup so we avoid any plan shenanigans by taking this value as soon as we leave the network generator resolution.

doudou (approved):

doudou:
I would feel better with a test that would fail on the old code.

jhonasiv:
> I would feel better with a test that would fail on the old code.

Done

wvmcastro (approved):

PR #493, opened 2025-04-09
fix: variable reference

seixasxbr:
### fix: variable reference
task_name dont exist in this context, the correct variable reference is deployed_task.name. This commit also brings more information to the error message

### DIFF
--- a/lib/syskit/process_managers/remote/process.rb
+++ b/lib/syskit/process_managers/remote/process.rb
@@ -73,7 +73,9 @@ module Syskit
                         mapped_name = mapped_name_of(deployed_task.name)
                         unless (ior = @ior_mappings[mapped_name])
                             raise IORNotRegisteredError,
-                                  "no IOR is registered for #{task_name}"
+                                  "no IOR is registered for #{deployed_task.name}. The "\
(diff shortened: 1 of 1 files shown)


lib/syskit/process_managers/remote/process.rb
```diff
@@ -73,7 +73,10 @@ def resolve_all_tasks
                         mapped_name = mapped_name_of(deployed_task.name)
                         unless (ior = @ior_mappings[mapped_name])
                             raise IORNotRegisteredError,
-                                  "no IOR is registered for #{task_name}"
+                                  "no IOR is registered for #{deployed_task.name}. The "\
+                                  "task deployment name on the process server differs "\
+                                  "to the task deployment name in syskit, "\
+                                  "#{deployed_task.name}"
```
doudou: The second part of the message is only a hint

The more general hint is "the deployment model used by the process server may differ from the one used by the syskit main instance"

doudou (approved):

jhonasiv (approved):

doudou:
There should have had a test added, rather obviously

PR #501, opened 2025-04-19
chore: cache result of dynamic port bindings based on matchers

doudou:
This was identified as a hotspot in syskit CPU usage on our systems.

We introduced a mechanism to avoid re-resolving queries for readers that have a
working reader already. This mechanism was actually not being triggered for the
very common case of matching a dataservice. This was because a data service's
actual port (the one that is used to create e.g. a reader) is not equal to the
one that is being matched. The matcher's current_selection_valid? method would
then not recognize its argument as valid.

Fix this by returning the "raw" port and letting DynamicPortBinding#update
resolving the actual port.

### DIFF
--- a/lib/syskit/dynamic_port_binding.rb
+++ b/lib/syskit/dynamic_port_binding.rb
@@ -31,6 +31,7 @@ module Syskit
         def initialize(model)
             @model = model
             @port_resolver = nil
+            @raw_resolved_port = nil
             @resolved_port = nil
             @accessor = nil
         end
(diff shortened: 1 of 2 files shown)


jhonasiv (approved):
Initially got confused by the PR's description about having go handle the finalize case, but I see its not in fact necessary by the second commit message. Maybe update the description?

wvmcastro (approved):

doudou:
> Maybe update the description?

Done. I thought it was needed but found out it was not.

PR #498, opened 2025-04-17
feat: add `update_and_restore_attr` helper

doudou:
The helper ensures that a modified attribute is properly restored to its old value, e.g. for global state (Conf, State or Syskit configuration parameters)

PR #502, opened 2025-04-23
feat: allow a remote manager to start disconnected / unavailable (on top of #498)

doudou:
On top of #498

This is only the remote manager side of the implementation. For now, a system which would be in this state would be unusable. The flag that is implemented in this PR is not functional.

### DIFF
--- a/lib/syskit/process_managers/remote/manager.rb
+++ b/lib/syskit/process_managers/remote/manager.rb
@@ -72,6 +72,7 @@ module Syskit
                 #   root for this client's loader
                 def initialize(
                     host = "localhost", port = DEFAULT_PORT,
+                    connect_timeout: 10,
                     response_timeout: 10,
                     root_loader: Orocos.default_loader,
                     register_on_name_server: true
@@ -79,7 +80,7 @@ module Syskit
                     @host = host
                     @port = port
                     @socket =
-                        begin TCPSocket.new(host, port)
+                        begin Socket.tcp(host, port, connect_timeout: connect_timeout)
                         rescue Errno::ECONNREFUSED => e
                             raise e.class,
                                   "cannot contact process server at " \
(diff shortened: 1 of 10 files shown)


lib/syskit/process_managers/remote/manager.rb
```diff
@@ -340,21 +295,97 @@ def join(deployment_name)
...
+                    @socket.write_nonblock(cmd)
+                    @socket.write_nonblock(Marshal.dump(args)) if args
+                end
+
+                def read_object(deadline:)
+                    validate_available
+
+                    timeout = [0, deadline - Roby.monotonic_time].max
+                    # This is no guarantee that Marshal.load won't block. Be careful
+                    unless select([@socket], [], [], timeout)
```
wvmcastro, line 412: What `select` returns in this case?

test/process_managers/test_remote.rb
```diff
@@ -53,6 +53,86 @@
...
+    end
+
+    describe "when accepting failed connections" do
+        before do
+            update_and_restore_attr(
+                Syskit.conf, :remote_process_managers_accept_failed_connections?, true
+            )
+        end
+
+        it "handles a server that allows connection but does not reply" do
```
wvmcastro, line 98: In what state a server would not reply?

wvmcastro (approved):
Just to be sure Remote::Manager is the client side handle for the server side Remote::Server, is that right?

doudou:
> Just to be sure Remote::Manager is the client side handle for the server side Remote::Server, is that right?

Right

test/process_managers/test_remote.rb
doudou: If a connection was lost, if it has a bug, if something goes wrong.

Essentially we have to assume that any read or write may be blocking at some point in time, and handle it correctly.

lib/syskit/process_managers/remote/manager.rb
doudou: The socket that has some event on it, or nil if things timed out. Since there's only one socket, the return value is treated as a boolean that means !timeout

PR #503, opened 2025-04-23
fix: warnings about deployment model being modified

doudou:
"Funnily" enough, the warning mentions the defined_deployed_task method which actually does not exist. Define it.

### DIFF
--- a/lib/syskit/models/deployment.rb
+++ b/lib/syskit/models/deployment.rb
@@ -107,6 +107,17 @@ module Syskit
                 klass
             end
 
+            # Define a new deployed task in this deployment model
+            #
+            # Mostly used in unit tests
+            #
+            # @return [OroGen::Spec::TaskDeployment]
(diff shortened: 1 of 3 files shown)


lib/syskit/models/deployment.rb
```diff
@@ -107,6 +107,14 @@ def new_submodel(name: nil, orogen_model: nil, **options, &block)
                 klass
             end
 
+            # Define a new deployed task in this deployment model
+            #
+            # Mostly used in unit tests
+            def define_deployed_task(name, model)
+                deployed_task = @orogen_model.task(name, model.orogen_model)
```
wvmcastro, line 116: `deployed_task` is being defined for what?

doudou: Was supposed to be returned. Fixed, and added unit tests.

wvmcastro (approved):

PR #511, opened 2025-05-24
fix: corner-case in template generation due to delayed arguments

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/343

Delayed arguments in general, and DelayedArgumentFromState in particular, may now be transformed into a static value during merge. This is great in the general state of things, but breaks templates.

This code disables the merge step for the delayed arguments, *except* in the case of default values, since default values do not have this behaviour *and* they are extremely common

### DIFF
--- a/test/test_instance_requirements.rb
+++ b/test/test_instance_requirements.rb
@@ -404,6 +404,101 @@ describe Syskit::InstanceRequirements do
         end
     end
 
+    describe "#compute_template" do
+        it "does not resolve delayed arguments" do
+            conf = Roby::ConfModel.new
+            conf.value = 10
+
+            task_m = Syskit::TaskContext.new_submodel
+            task_m.argument :arg, default: Roby.from_state(conf).value
+            cmp_m = Syskit::Composition.new_submodel
+            cmp_m.add task_m, as: "test"
+
+            ir = cmp_m.to_instance_requirements
+            cmp = ir.compute_template.root_task
+            assert_kind_of Roby::DelayedArgumentFromState,
+                           cmp.test_child.arguments.raw_get(:arg)
+        end
+
+        it "applies merges" do
+            task_m = Syskit::TaskContext.new_submodel
+            task_m.argument :arg
+            cmp_m = Syskit::Composition.new_submodel
+            cmp_m.add task_m, as: "task0"
+            cmp_m.add task_m, as: "task1"
+
+            ir = cmp_m.use(
+                "task0" => task_m.with_arguments(arg: 10),
+                "task1" => task_m.with_arguments(arg: 10)
+            )
+            cmp = ir.compute_template.root_task
+            assert_same cmp.task0_child, cmp.task1_child
+        end
+
+        it "does not apply merges if there are tasks with delayed arguments" do
+            conf = Roby::ConfModel.new
+            conf.value = 10
+
+            task_m = Syskit::TaskContext.new_submodel
+            task_m.argument :arg
+            cmp_m = Syskit::Composition.new_submodel
+            cmp_m.add task_m, as: "task0"
+            cmp_m.add task_m, as: "task1"
+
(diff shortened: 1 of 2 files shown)


test/test_instance_requirements.rb
```diff
@@ -404,6 +404,94 @@
...
+
+            ir = cmp_m.use(
+                "task0" => task_m.with_arguments(arg: 20), "task1" => task_m
+            )
+            cmp = ir.compute_template.root_task
+            assert_same cmp.task0_child, cmp.task1_child
+            assert_equal 20, cmp.task0_child.arg
+        end
+
+        it "applies merges if there are from(:parent_task) patterns" do
```
wvmcastro, line 475: where is the merge here?

test/test_instance_requirements.rb
```diff
@@ -404,6 +404,94 @@
...
+            cmp_m = Syskit::Composition.new_submodel
+            cmp_m.add task_m, as: "test"
+
+            ir = cmp_m.to_instance_requirements
+            cmp = ir.compute_template.root_task
+            assert_kind_of Roby::DelayedArgumentFromState,
+                           cmp.test_child.arguments.raw_get(:arg)
+        end
+
+        it "applies merges" do
```
wvmcastro, line 423:
```suggestion
        it "applies merges for static value arguments" do
```

test/test_instance_requirements.rb
```diff
@@ -404,6 +404,94 @@
...
+            )
+            cmp = ir.compute_template.root_task
+            refute_equal cmp.task0_child, cmp.task1_child
+            assert_kind_of Roby::DelayedArgumentFromState,
+                           cmp.task0_child.arguments.raw_get(:arg)
+            assert_kind_of Roby::DelayedArgumentFromState,
+                           cmp.task1_child.arguments.raw_get(:arg)
+        end
+
+        it "applies merges if the only delayed arguments are default arguments" do
```
wvmcastro, line 460: would it apply even if the delayed argument was a state?

test/test_instance_requirements.rb
```diff
@@ -404,6 +404,94 @@
...
+            task_m.argument :arg
+            cmp_m = Syskit::Composition.new_submodel do
+                argument :arg
+
+                add(task_m, as: "task0")
+                    .with_arguments(arg: from(:parent_task).arg)
+                add task_m, as: "task1"
+            end
+
+            ir = cmp_m.use("task1" => task_m.with_arguments(arg: 20))
```
jhonasiv: missing a `.with_arguments` for cmp_m to set from(:parent_task).arg to 20 as well so it can be merged

test/test_instance_requirements.rb
doudou: I did not want to say for what because the list is big. Preferred enumerating the exceptions below.

test/test_instance_requirements.rb
doudou: I think you guys are up to something ... de-facto, this tests that the from(:parent_task) is not merged with the static value.

test/test_instance_requirements.rb
doudou: That's a ... very good question. It should actually not merge in this particular instance.

test/test_instance_requirements.rb
doudou: This is not allowed. The code resolves `default: delayed_argument` into `delayed_argument` instead of wrapping it with a `Roby::DefaultArgument`

test/test_instance_requirements.rb
doudou: `from(:parent_task)` should not merge (because of its own rules). Added another task with a static argument to check that the merge step happens.

wvmcastro (approved):

jhonasiv (approved):

PR #513, opened 2025-05-26
fix: avoid considering tasks without a deployment for conflicting deployment allocation

jhonasiv:
They cant be conflicting if they have no deployment after all...

### DIFF
--- a/lib/syskit/network_generation/system_network_generator.rb
+++ b/lib/syskit/network_generation/system_network_generator.rb
@@ -479,8 +479,9 @@ module Syskit
                 deployment_to_task_map = plan.find_local_tasks(Syskit::TaskContext)
                                              .group_by(&:orocos_name)
 
-                using_same_deployment = deployment_to_task_map.select do |_, tasks|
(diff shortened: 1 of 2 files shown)


wvmcastro (approved):

lib/syskit/network_generation/system_network_generator.rb
```diff
@@ -479,7 +479,10 @@ def self.verify_all_deployments_are_unique(
                 deployment_to_task_map = plan.find_local_tasks(Syskit::TaskContext)
                                              .group_by(&:orocos_name)
 
-                using_same_deployment = deployment_to_task_map.select do |_, tasks|
+                using_same_deployment = deployment_to_task_map.select do |name, tasks|
+                    # There cant be a conflict between tasks that have no deployment
+                    next unless name
+
                     tasks.size > 1
```
doudou: I would personally prefer just `name && tasks.size > 1`, or doing `deployment_to_task_map.delete(nil)` before the select ...

doudou (approved):

PR #514, opened 2025-05-29
fix: support multiple exceptions on ProfileAssertionFailed

jhonasiv:
When the capture_errors_during_network_resolution flag is on, the ProfileAssertionFailed will receive an array of exceptions (even if its a single exception). At first glance it wasnt an issue, but Roby.display_exception expects that the array is flattened, so it would raise whenever trying to display exceptions with a ProfileAssertionFailed. 

I prefer to solve this than flattening the exceptions on Roby#display_exceptions (as in https://github.com/rock-core/tools-roby/pull/342), which would hide unintended behaviors.

### DIFF
--- a/lib/syskit/test/profile_assertions.rb
+++ b/lib/syskit/test/profile_assertions.rb
@@ -15,10 +15,13 @@ module Syskit
             class ProfileAssertionFailed < Roby::ExceptionBase
                 attr_reader :actions
 
-                def initialize(assertion_name, actions, original_error)
+                def initialize(assertion_name, actions, original_errors)
                     @assertion_name = assertion_name
                     @actions = Array(actions)
-                    super([original_error])
+                    # When capture errors is active, its possible to receive an array of
+                    # exceptions instead of a single one.
+                    original_errors = Array(original_errors)
+                    super(original_errors)
                 end
 
                 def pretty_print(pp)
--- a/test/test/test_profile_assertions.rb
+++ b/test/test/test_profile_assertions.rb
@@ -572,6 +572,24 @@ module Syskit
                     )
                 end
 
+                it "accepts failing with more than a single exception and can display " \
+                   "the exception" do
+                    skip unless Syskit.conf.capture_errors_during_network_resolution?
+
+                    @test_profile.define "test", @cmp_m
+
+                    e = assert_raises(ProfileAssertions::ProfileAssertionFailed) do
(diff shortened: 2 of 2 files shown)


lib/syskit/test/profile_assertions.rb
```diff
@@ -15,10 +15,16 @@ module ProfileAssertions
...
                     @assertion_name = assertion_name
                     @actions = Array(actions)
-                    super([original_error])
+                    # When capture errors is active, its possible to receive an array of
+                    # exceptions instead of a single one.
+                    unless original_errors.kind_of? Array
+                        original_errors = [original_errors]
+                    end
+                    super(original_errors)
+                    # super([original_errors])
```
doudou: remove ...

lib/syskit/test/profile_assertions.rb
```diff
@@ -343,7 +349,7 @@ def assert_can_instanciate_together(*actions)
                 )
             rescue NetworkGeneration::PartialNetworkResolution => e
                 raise ProfileAssertionFailed.new("configure together", actions,
-                                                 e.original_exceptions)
+                                                 *e.original_exceptions)
```
doudou: Adding the splat will cause a "wrong number of arguments" errors if there is more than one original_exceptions. `initialize` expects only 3 arguments.

Keep initialize as it is and remove the splat here.

jhonasiv: That wasnt supposed to be in the PR... I was experimenting a bit and it went through

lib/syskit/test/profile_assertions.rb
```diff
@@ -15,10 +15,15 @@ module ProfileAssertions
...
-                def initialize(assertion_name, actions, original_error)
+                def initialize(assertion_name, actions, original_errors)
                     @assertion_name = assertion_name
                     @actions = Array(actions)
-                    super([original_error])
+                    # When capture errors is active, its possible to receive an array of
+                    # exceptions instead of a single one.
+                    unless original_errors.kind_of? Array
+                        original_errors = [original_errors]
+                    end
```
doudou: FYI, `Array()` is the Ruby canonical conversion method, that converts anything-to-array (but keeps arrays as-is), i.e.

```
original_errors = Array(original_errors)
```

doudou (approved):

wvmcastro (approved):

PR #520, opened 2025-07-23
fix: make sure syskit_join_current_resolution does not start a new resolution

doudou:
I really can't find why this was done. My guess is that the new call was meant to apply the result of the current resolution. But then, it would also start a new one, which is rather obviously NOT expected.

This fixes weird behaviors during tests, now that we call this explicitly in teardown to CANCEL the current async resolution ...

### DIFF
--- a/lib/syskit/runtime/apply_requirement_modifications.rb
+++ b/lib/syskit/runtime/apply_requirement_modifications.rb
@@ -85,7 +85,7 @@ module Syskit
                 rescue Concurrent::CancelledOperationError # rubocop:disable Lint/SuppressedException
                 end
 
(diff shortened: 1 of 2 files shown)


jhonasiv (approved):
I approve the change. 

About the previous behavior, wouldnt the `#apply_requirement_modifications` call go in an early return before actually starting a new resolution? It treats running obsolete resolutions as well as avoids starting a new one is there is an active resolution. My question is in what circumstance this would cause the issues in tests you mentioned?

doudou:
True if there is a resolution running. But `syskit_joint_current_resolution` can actually be called without a resolution running (it is documented as such).

In addition, I really wonder about the early return thing ... not sure why it exists.

PR #521, opened 2025-07-25
fix: dont fail tasks not in the resolution

jhonasiv:
This only happened on the non-error capture code path, but still shouldnt. The requirement tasks should always be the tasks included in the resolution

### DIFF
--- a/test/runtime/test_apply_requirement_modifications.rb
+++ b/test/runtime/test_apply_requirement_modifications.rb
@@ -79,6 +79,59 @@ module Syskit
             end
         end
 
+        describe ".syskit_apply_async_resolution_results" do
+            it "ignores instance requirement tasks added to the plan after the " \
+               "resolution was started and would succeded without errors" do
+                cmp_m = Composition.new_submodel
+                cmp = plan.add_permanent_task(cmp_m.to_instance_requirements.as_plan)
+                requirement_task = cmp.planning_task
+
+                execute { requirement_task.start! }
+                execute { Runtime.apply_requirement_modifications(plan) }
+                plan.syskit_current_resolution.future.value
+
+                # This task should fail if listed for deployment as it has no defined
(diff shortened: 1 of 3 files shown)


doudou (approved):

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -79,6 +79,56 @@ module Runtime
...
+                cmp = plan.add_permanent_task(@cmp_m.to_instance_requirements.as_plan)
+                @requirement_task = cmp.planning_task
+            end
+
+            it "ignores instance requirement tasks added to the plan after the " \
+                "resolution was started" do
+                requirement_task = @requirement_task
+                execute { requirement_task.start! }
+                execute { Runtime.apply_requirement_modifications(plan) }
+                plan.syskit_current_resolution.future.value
```
wvmcastro, line 91: what does this call do?

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -79,6 +79,56 @@ module Runtime
             end
         end
 
+        describe ".syskit_apply_async_resolution_results" do
+            before do
+                @cmp_m = Composition.new_submodel
+                cmp = plan.add_permanent_task(@cmp_m.to_instance_requirements.as_plan)
+                @requirement_task = cmp.planning_task
```
wvmcastro: Looks like these are used just for the first test, therefore I think it should be scoped accordingly

test/runtime/test_apply_requirement_modifications.rb
jhonasiv: its purpose here is to block until the future is fulfilled, it also returns the value

wvmcastro (approved):
we've talked in person and my doubts are clarified

PR #522, opened 2025-07-31
feat: override model instead of klass

jhonasiv:
The model is more informative than the instance, and in hindsight is the better choice to be overriden. For example, by overriding the model we can change the type of the reader. Keep in mind whoever overrides the model probably still wants to implement a new instanciated version of it.  Do so by overriding the `#instanciate` method

### DIFF
--- a/test/test_component.rb
+++ b/test/test_component.rb
@@ -691,6 +691,22 @@ describe Syskit::Component do
             expect_execution.to { achieve { reader.connected? } }
             assert_equal task.out_port, reader.resolved_accessor.port
         end
+
+        it "overrides BoundOutputReader with another model" do
+            injected = Class.new Syskit::Models::DynamicPortBinding::BoundOutputReader do
+                def type
+                    String
+                end
+            end
+
+            @support_task_m.data_reader(
+                @task_m.match.running.out_port, as: "test", model: injected
+            )
+            reader = @support_task_m.test_reader
+
+            assert reader.kind_of? injected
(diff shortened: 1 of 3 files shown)


test/test_component.rb
```diff
@@ -691,6 +691,50 @@
...
+            @support_task_m.data_reader(
+                @task_m.match.running.out_port, as: "test", model: injected
+            )
+            reader = @support_task_m.test_reader
+
+            assert reader.kind_of? injected
+            assert_equal String, reader.type
+        end
+
+        it "allows an injected model to override the reader instance's class" do
```
jhonasiv: @doudou this test represents the mechanism to indicate the class when instantiating the output reader. IMO it is fine, this requires the model to know the class of the instance it generates, which makes sense. What do you think about it?

doudou:
> Keep in mind whoever overrides the model probably still wants to implement a new instanciated version of it. Do so by setting the klass.

This sentence ~~makes no sense.~~ was very confusing.

test/test_component.rb
doudou: Strong dislike.

If someone wants to change the data reader instance, that someone has to overload `#instanciate`

doudou (commented):
I would just remove the test about `@klass`. I really believe one should simply override `instanciate` in that case.

doudou (approved):

wvmcastro (approved):

PR #524, opened 2025-08-05
fix: stay aware of pending forced resolution when applying modifications

jhonasiv:


lib/syskit/runtime/apply_requirement_modifications.rb
```diff
@@ -133,22 +137,26 @@ def self.apply_requirement_modifications(
             if plan.syskit_has_async_resolution?
                 # We're already running a resolution, make sure it is not
                 # obsolete
-                valid = !force && plan.syskit_valid_async_resolution?(
+                obsolete = force || !plan.syskit_valid_async_resolution?(
```
doudou: I preferred the version I saw on your computer, that is the one that separated `force` from `obsolete`

lib/syskit/runtime/apply_requirement_modifications.rb
```diff
@@ -133,22 +137,26 @@ def self.apply_requirement_modifications(
...
             # NOTE: this is NOT the 'else' clause from above. The if ... block
             # above might resolve the current async resolution
-            return if plan.syskit_has_async_resolution?
+            if plan.syskit_has_async_resolution?
+                @pending_forced_resolution = force
+                return
+            end
+
+            force ||= plan.syskit_pending_forced_resolution?
+            @pending_forced_resolution = false
```
doudou: Shouldn't that be `plan.pending_forced_resolution =` ?

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -304,6 +304,190 @@ module Runtime
...
+                    requirement_task =
+                        plan.add_permanent_task(t.to_instance_requirements.as_plan)
+                    requirement_task.planning_task
+                end
+                execute { reqs.first.start! }
+                execute { Runtime.apply_requirement_modifications(plan) }
+                execute { reqs[1].start! }
+                execute { Runtime.apply_requirement_modifications(plan) }
+                execute { reqs[2].start! }
+                execute { Runtime.apply_requirement_modifications(plan) }
```
wvmcastro:
```ruby
reqs.each do |req|
    execute { req.start! }
    execute { Runtime.apply_requirement_modifications(plan) }
end
```

test/runtime/test_apply_requirement_modifications.rb
wvmcastro, line 1: I think the helper
```ruby
def add_instace_requirement_to_plan
    cmp_m = Composition.new_submodel
    req_task = plan.add_permanent_task(cmp_m.to_instance_requirements.as_plan)
    req_task.planning_task
end
```
would make the tests less verbose and more straight forward

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -304,6 +304,190 @@ module Runtime
...
+                execute { requirement_task.start! }
+                execute { Runtime.apply_requirement_modifications(plan, force: true) }
+                assert plan.syskit_has_async_resolution?
+                FlexMock.use(plan) do |mock|
+                    mock.should_receive(syskit_finished_async_resolution?: false)
+                    execute { Runtime.apply_requirement_modifications(plan, force: true) }
+                end
+                assert plan.syskit_current_resolution.cancelled?
+                execute { Runtime.apply_requirement_modifications(plan, force: true) }
+                execute { Runtime.apply_requirement_modifications(plan, force: true) }
```
wvmcastro: It is not clear to me why twice?

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -304,6 +304,202 @@ module Runtime
                 assert other_requirement_task.failed?
             end
 
+            it "resolves all requirements that were added to the plan in successive " \
+               "#apply_requirement_modifications calls" do
```
wvmcastro, line 308: was this test added for completeness sake or it was a bug?

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -304,6 +304,202 @@ module Runtime
...
+                assert plan.syskit_has_async_resolution?
+
+                execute { plan.syskit_join_current_resolution }
+                reqs.each do |req|
+                    assert req.resolution_success?
+                end
+            end
+
+            it "while using force, cancels pending resolutions and executes the last " \
+               "one" do
```
wvmcastro, line 330: I would put this and the below tests under a `describe "force"` namespace

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -304,6 +304,202 @@ module Runtime
...
+                end
+                assert plan.syskit_current_resolution.cancelled?
+                execute { Runtime.apply_requirement_modifications(plan, force: true) }
+                execute { Runtime.apply_requirement_modifications(plan, force: true) }
+                assert plan.syskit_has_async_resolution?
+                refute plan.syskit_current_resolution.cancelled?
+            end
+
+            it "triggers a new resolution once with force: false when a force was " \
+               "used while a resolution was pending" do
```
wvmcastro, line 354:
```suggestion
            it "always triggers a new resolution while there is a pending forced one" do
```

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -304,6 +304,202 @@ module Runtime
...
+                requirement_task = requirement_task.planning_task
+                execute { requirement_task.start! }
+
+                # Finish one resolution, we're checking behaviour related to the force
+                # flag
+                execute { Runtime.apply_requirement_modifications(plan) }
+                begin
+                    plan.syskit_current_resolution.join
+                rescue Concurrent::CancelledOperationError # rubocop:disable Lint/SuppressedException
+                end
```
wvmcastro: I don't get this rescue, why the `join` would raise?

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -304,6 +304,202 @@ module Runtime
...
+                execute { Runtime.apply_requirement_modifications(plan) }
+                refute plan.syskit_has_async_resolution?
+                assert plan.syskit_pending_forced_resolution?
+
+                # Triggers new resolution because of the last 'force'
+                execute { Runtime.apply_requirement_modifications(plan) }
```
wvmcastro, line 393: I expected that the previous `apply_requirement_modifications` call would trigger a new resolution bacause of force, already. Why is not the case?

test/runtime/test_apply_requirement_modifications.rb
jhonasiv: because the operation was cancelled?! Dont get confused with `syskit_join_current_resolution`, which does this exact operation with rescue AND applies the resolution results

test/runtime/test_apply_requirement_modifications.rb
jhonasiv: The previous call detects that the resolution was finished and, but doesnt start a new one (starting a new resolution done in this call)

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -304,6 +304,202 @@ module Runtime
...
+                requirement_task = requirement_task.planning_task
+                execute { requirement_task.start! }
+                execute { Runtime.apply_requirement_modifications(plan, force: true) }
+                assert plan.syskit_has_async_resolution?
+                FlexMock.use(plan) do |mock|
+                    mock.should_receive(syskit_finished_async_resolution?: false)
+                    execute { Runtime.apply_requirement_modifications(plan, force: true) }
+                end
+                assert plan.syskit_current_resolution.cancelled?
+                execute { Runtime.apply_requirement_modifications(plan, force: true) }
```
doudou, line 346: You need to join cancelled resolutions to create new heisenbugs. I suggest doing a complete read with only this in mind ...

test/runtime/test_apply_requirement_modifications.rb
doudou: Since you'll need this more than once ... a helper method would be a good idea ;-)

test/runtime/test_apply_requirement_modifications.rb
```diff
@@ -304,6 +304,202 @@ module Runtime
...
+                    mock.should_receive(syskit_finished_async_resolution?: false)
+                    # Cancels resolution, but does nothing else as
+                    # there is a pending unfinished resolution
+                    #
+                    # The test is to check that the `force` flag is remembered and will
+                    # be used the next time
+                    execute { Runtime.apply_requirement_modifications(plan, force: true) }
+                    execute do
+                        Runtime.apply_requirement_modifications(plan, force: false)
+                    end
```
doudou, line 439: ? I would try to have as little calls to apply_* as possible. If there are other conditions you want to check, make more tests for them,

test/runtime/test_apply_requirement_modifications.rb
jhonasiv: for completeness sake

test/runtime/test_apply_requirement_modifications.rb
jhonasiv: meant to do that but it got through without me realization

test/runtime/test_apply_requirement_modifications.rb
jhonasiv: this test doesnt check all the possible calls to apply requirements modifications, so I cant say that it "always" does something

test/runtime/test_apply_requirement_modifications.rb
jhonasiv: force is a very poor description, also I prefer not having to deal with the extra indentation in this case

wvmcastro (approved):

doudou (approved):

PR #528, opened 2025-09-08
chore: add conf section helper methods for validation and tests

jhonasiv:


wvmcastro (approved):

lib/syskit/task_configuration_manager.rb
```diff
@@ -118,5 +118,13 @@ def load_from_yaml(path, cache_dir: self.cache_dir)
             FileUtils.mkdir_p self.cache_dir
             super
         end
+
+        # Helpers method that checks if the given conf section is valid for the task model
+        def validate_conf_section(section)
+            return if has_section?(section)
+
+            raise ArgumentError,
```
wvmcastro: Giving a second thought, I think it would be better to raise the existing `SectionNotFound` exception

doudou (approved):

PR #532, opened 2025-11-17
fix: put a task in quarantine if it times out on stop

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-orogen_syskit_tests/pull/9

We have had a few instances of tasks that would be STOPPED on the C++ side after syskit attempted to interrupt them, but for which Syskit assumed the task was still running.

Make sure we time out in that case and quarantine the task (which under current configuration should actually kill the deployment)

### DIFF
--- a/test/live/test_blocking_hooks.rb
+++ b/test/live/test_blocking_hooks.rb
@@ -462,6 +462,33 @@ module Syskit
             end
         end
 
+        describe "stop transition timeout" do
+            attr_reader :deployment
+
+            before do
+                @task = syskit_deploy_configure_and_start(
+                    OroGen.orogen_syskit_tests.Empty
+                          .deployed_as("stop-transition-timeout")
+                )
+                @deployment = @task.execution_agent
+                @__stop_transition_timeout = Syskit.conf.stop_transition_timeout
+            end
+
+            after do
+                cleanup_running_tasks
+                Syskit.conf.stop_transition_timeout = @__stop_transition_timeout
+            end
+
+            it "detects a task that should have stopped but has not" do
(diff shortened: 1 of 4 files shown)


wvmcastro (commented):
I didn't find where `@stop_transition_deadline` is computed

doudou:
> I didn't find where `@stop_transition_deadline` is computed

Good catch ... Got lost when I removed the debug stuff.

At least the new test catches it ... so good test :)

wvmcastro:
> > I didn't find where `@stop_transition_deadline` is computed
> 
> Good catch ... Got lost when I removed the debug stuff.
> 
> At least the new test catches it ... so good test :)

The test only covers that the task is quarantined. It does not test if the "quarantine deadline" was respected, which it did not

test/live/test_blocking_hooks.rb
```diff
@@ -462,6 +462,33 @@ def task.configure
...
+                Syskit.conf.stop_transition_timeout = @__stop_transition_timeout
+            end
+
+            it "detects a task that should have stopped but has not" do
+                Syskit.conf.stop_transition_timeout = 1
+                @task.should_receive(:stop_orocos_task)
+                tic = Time.now
+                expect_execution { @task.stop! }
+                    .to_quarantine(@task)
+                assert (Time.now - tic) > 1
```
wvmcastro, line 488: I think lately we are moving to 
```suggestion
                assert_operator Time.now - tic, :>, 1
```

but, to be clear, it is not a big deal for me and is not in the code guidelines whatsoever

wvmcastro (approved):

jhonasiv (approved):

PR #533, opened 2025-11-18
chore: move the thread-specific logic for the async resolution into its own class

doudou:
The objective is to have other execution policies than multithreading while keeping the existing logic available

### DIFF
--- a/lib/syskit/network_generation/async.rb
+++ b/lib/syskit/network_generation/async.rb
@@ -2,146 +2,109 @@
 
 module Syskit
     module NetworkGeneration
-        # A partially asynchronous requirement resolver built on top of {Engine}
+        # Base class and interface definition for the async resolver used
+        # by {Runtime.apply_requirements_modification}
         class Async
             extend Logger::Hierarchy
             include Logger::Hierarchy
             include Roby::DRoby::EventLogging
 
-            # The target plan
-            attr_reader :plan
-
-            # The {Roby::DRoby::EventLogger} used to log timings
-            attr_reader :event_logger
-
-            # The thread pool (or, really, any of Concurrent executor)
-            attr_reader :thread_pool
-
-            # The future that does the async work
-            #
-            # It is created by {#start}
-            #
-            # @return [Resolution]
-            attr_reader :future
-
-            def initialize(plan, event_logger: plan.event_logger,
-                thread_pool: Concurrent::CachedThreadPool.new)
-                @plan = plan
-                @event_logger = event_logger
-                @thread_pool = thread_pool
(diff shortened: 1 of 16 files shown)


lib/syskit/network_generation/async_threaded.rb
```diff
@@ -0,0 +1,124 @@
...
+
+            def initialize(
+                plan, requirement_tasks,
+                event_logger: plan.event_logger, **resolver_options
+            )
+                super
+
+                @thread_pool = Concurrent::CachedThreadPool.new
+
+                @keepalive = Roby::Transaction.new(plan)
```
wvmcastro, line 45: why the transaction perform a _"keepalive"_ role? Why it is not just _work_plan_ ?

lib/syskit/network_generation/async_threaded.rb
```diff
@@ -0,0 +1,124 @@
...
+
+                cancel if !cancelled? && !valid?(requirement_tasks)
+
+                return unless network_generation_complete?
+
+                apply_complete_network_generation
+            end
+
+            def apply_complete_network_generation
+                running_requirement_tasks = @requirement_tasks.find_all(&:running?)
```
wvmcastro: where `@requirement_tasks` is being initialized? I couldn't find

lib/syskit/network_generation/async.rb
wvmcastro, line 1: since `Async` also defines an interface, I would like to see a test in the sense

```ruby
Async.each_subclass { |c| implements_async_interface(c) }
```

lib/syskit/network_generation/async_threaded.rb
doudou: the transaction keeps alive all components while the system is being re-deployed, i.e. protects them against garbage collection to avoid stopping a component that was used by some (now useless) instance requirements, but will be re-used by a new requirement.

lib/syskit/network_generation/async.rb
doudou: I don't think this particular bit of code warrants such a work. The refactoring is meant as a way forward to drop one implementation and replace it with another, not as a generic interface that anyone could want to implement.

We are checking that the only user of that interface (apply_requirements) is happy with it, I think that's more than enough. In the PR that implements the fiber-based async, I check that it is happy with both.

lib/syskit/network_generation/async_threaded.rb
wvmcastro: I haven't seen, but it was at the base class all long

wvmcastro (approved):

jhonasiv (approved):

PR #535, opened 2025-11-18
feat: implement fiber-based async resolution (on top of #533)

doudou:
On top of #533 

"Fibers" are cooperative concurrency primitives in Ruby. The thread-based implementation clearly does not work - the main thread will seldom interact with "the outside world" while syskit is deploying, which is a major source of issues.


jhonasiv (approved):

wvmcastro (approved):

doudou:
As discussed, I'm merging this into #533 and will merge 533 when CI passes.

PR #539, opened 2025-12-05
feat: implement the lazy-deploy mode

doudou:
The current network generation + adaptation algorithm currently:
- generates a fully formed network in network generation, which means:
  * generating the task tree (compositions + task contexts)
  * replacing the task contexts by deployed tasks (via early deploy or late deploy)
- adapt the running network to match the generated network
  * which means, when running components are reused, replacing *again* tasks with the proxy of the deployed tasks from the plan

With early deploy, it means instantiating more deployments and deployed tasks than the final number of task contexts (because early deploy happens before merge), to then - in practice - replace most of them by the tasks from the plan (assuming we don't change the whole system all the time).

The late deploy mode tries to avoid all that wasted energy. Deployment in system network generation is now only setting the `orocos_name` field. We generate new deployed tasks only during the plan adaptation, and only when needed.


doudou:
@wvmcastro @jhonasiv I *strongly* recommend to first do a review commit-by-commit. I refactored Engine to extract the runtime adaptation bits, which is a good thing, but made the mistake to not spend the effort to have a clean PR with just this.

Looking commit-by-commit would allow you to see what actually is being changed on that part of the code between the original code and the one that is "ready" for lazy deployt

lib/syskit/network_generation/system_network_deployer.rb
```diff
@@ -250,11 +275,62 @@ def verify_all_tasks_deployed(error_handler: RaiseErrorHandler.new)
...
+                    plan, default_deployment_group, error_handler: error_handler
+                )
+            end
+
+            # @see #verify_all_tasks_lazily_deployed
+            #
+            # @param [Component=>DeploymentGroup] deployment_groups which
+            #   deployment groups has been used for which task. This is used
+            #   to generate the error messages when needed.
+            def self.verify_all_tasks_lazily_deployed(
```
jhonasiv: I dont see the use of this methods. the changes you made on `verify_all-tasks_deployed` is enough IMO

lib/syskit/network_generation/engine.rb
```diff
@@ -127,7 +127,7 @@ def compute_deployed_network(
                         resolution_control: @resolution_control
                     )
 
-                    deployer.deploy(
+                    @used_deployments, = deployer.deploy(
```
jhonasiv: Why not forward the used deployments up until the calling places of `apply_system_network_to_plan` (the `resolve` method and async AFAIK)? It feels cleaner even if I cant find another behavior problem that comes from the way you implemented it

lib/syskit/network_generation/merge_solver.rb
```diff
@@ -235,8 +235,9 @@ def may_merge_components?(merged_task, task)
 
             # Tests whether task.merge(target_task) is a valid operation
             #
-            # @param [Syskit::TaskContext] task
-            # @param [Syskit::TaskContext] target_task
+            # @param [Syskit::TaskContext] merged_task the task that we will merge into
+            #   the other one. This task disappears after the merge
```
wvmcastro: "this task disapperars after the merge" is gold

lib/syskit/network_generation/system_network_deployer.rb
```diff
@@ -250,11 +285,62 @@ def verify_all_tasks_deployed(error_handler: RaiseErrorHandler.new)
...
+                error_handler: RaiseErrorHandler.new, lazy: false
+            )
+                query = plan.find_local_tasks(TaskContext)
+                            .not_finished.not_abstract
+                not_deployed =
+                    if lazy
+                        query.find_all { !_1.orocos_name }
+                    else
+                        query.find_all { !_1.execution_agent }
+                    end
```
wvmcastro, line 355: Registering what we've talked in person: Create a `#deployed?` predicate to abstract what being deployed means

wvmcastro (commented):
I think you could reuse the computed _used_deployments_ between `#compute_system_network` and `#compute_deployed_network` at `#resolve_system_network` when early deploy is on.

lib/syskit/network_generation/system_network_deployer.rb
doudou: removed

lib/syskit/network_generation/engine.rb
doudou: The amount of things we would need to "remove" the state from Engine itself is ... big actually.

I think we'll need sometimes to either do all of it or none of it, but right now it is 80% inside the task and 20% as argument... I don't think there's a "winner" w.r.t. being consistent.

lib/syskit/network_generation/system_network_deployer.rb
doudou: Done.

jhonasiv (approved):

PR #541, opened 2025-12-15
chore: use the `cache` flag for the data_service inherited attribute (on top of #542)

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-metaruby/pull/26
- [ ] https://github.com/rock-core/tools-syskit-log/pull/65
- [ ] https://github.com/rock-core/tools-roby/pull/377

On top of #542

### DIFF
--- a/lib/syskit/models/component.rb
+++ b/lib/syskit/models/component.rb
@@ -159,7 +159,7 @@ module Syskit
                     return enum_for(:each_slave_data_service, master_service)
                 end
 
-                each_data_service(nil) do |_name, service|
+                each_data_service do |_name, service|
                     next unless (m = service.master)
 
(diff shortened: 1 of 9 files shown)


lib/syskit/models/component.rb
```diff
@@ -55,8 +55,8 @@ def can_use_template?
 
             def clear_model
                 super
-                data_services.clear
-                dynamic_services.clear
+                data_services_clear
+                dynamic_services_clear
```
jhonasiv, line 59: couldnt find references to these methods, where are they?

doudou: Defined by `inherited_attributes`

jhonasiv (approved):

jhonasiv (approved):

PR #545, opened 2026-02-26
chore: rescue file system handling exceptions at ensure free space

wvmcastro:
catch exceptions when
1. listing directory
2. deleting files

### DIFF
--- a/lib/syskit/cli/log_runtime_archive.rb
+++ b/lib/syskit/cli/log_runtime_archive.rb
@@ -101,8 +101,8 @@ module Syskit
 
                 return true if available_space > free_space_low_limit
 
+                files = list_files(directory).sort_by(&:mtime)
                 until available_space >= free_space_delete_until
-                    files = directory.each_child.select(&:file?)
                     if files.empty?
                         @logger.warn(
                             "The available space (#{available_space}) is below the " \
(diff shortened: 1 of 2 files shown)


doudou (commented):
This really does not solve the problem. If there is an exception, the free space tool will not die anymore, but won't free any space anyways.

We can have exceptions that we need to ignore everywhere there is an interaction with the filesystem, which is essentially:
- listing the directories (we don't have permission for that)
- deleting files (we can list but can't delete)

I think the main loop in ensure_free_space should not have a begin/rescue. I would create a method to list files, that handles exceptions and create a method to delete them that also gracefully handles exceptions.

lib/syskit/cli/log_runtime_archive.rb
```diff
@@ -113,18 +113,40 @@ def ensure_free_space(
...
+            end
+
+            # Deletes a file and returns the freed space. Returns 0 on exceptions
+            def delete_file(file)
+                size = file.size
+                file.unlink
+                size
+            rescue ::SystemCallError => e
+                @logger.error(e)
+                0
```
doudou: I would prefer `nil`. It is weird to return `0` on exception.

doudou (approved):

PR #548, opened 2026-03-10
fix: ensure to remove tasks with errors from used deployments

jhonasiv:
There were two different scenarios where we got the same kind of error, which looked like this:

```
NoMethodError: undefined method `[]' for nil:NilClass
    /home/jhonas/dev/wetpaint/tools/roby/lib/roby/relations/space.rb:340:in `block (2 levels) in relation'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/runtime_network_adaptation.rb:454:in `adapt_existing_deployment'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/runtime_network_adaptation.rb:177:in `handle_required_deployment'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/runtime_network_adaptation.rb:129:in `block in finalize_used_deployments'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/runtime_network_adaptation.rb:127:in `each'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/runtime_network_adaptation.rb:127:in `finalize_used_deployments'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/runtime_network_adaptation.rb:66:in `finalize_deployed_tasks'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/runtime_network_adaptation.rb:40:in `apply'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/engine.rb:186:in `block in apply_deployed_network_to_plan'
    /home/jhonas/dev/wetpaint/tools/roby/lib/roby/event_logging/mixin.rb:43:in `log_timepoint_group'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/engine.rb:179:in `apply_deployed_network_to_plan'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/engine.rb:613:in `block in apply_system_network_to_plan'
    /home/jhonas/dev/wetpaint/tools/roby/lib/roby/event_logging/mixin.rb:43:in `log_timepoint_group'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/engine.rb:612:in `apply_system_network_to_plan'
    /home/jhonas/dev/wetpaint/tools/syskit/lib/syskit/network_generation/engine.rb:587:in `resolve'
    /home/jhonas/dev/wetpaint/tools/syskit/test/network_generation/test_engine.rb:737:in `block (4 levels) in <module:NetworkGeneration>'
```

That happened because we were removing tasks from the plan due to a resolution error during the #apply_deployed_network_to_plan step, but we never removed them from the used_deployments, therefore the system was still trying to solve for tasks that were not in the plan anymore.

### DIFF
--- a/lib/syskit/network_generation/engine.rb
+++ b/lib/syskit/network_generation/engine.rb
@@ -133,8 +133,12 @@ module Syskit
                         lazy: lazy_deploy
                     )
                     @used_deployments = @used_deployments.merge(used_deployments)
-                    resolution_errors = error_handler.process_failures(
-                        required_instances, cleanup_failed_tasks: true
+                    @used_deployments.transform_keys! do |task|
+                        merge_solver.replacement_for(task)
+                    end
+                    resolution_errors, @used_deployments = process_failures(
+                        error_handler, required_instances, @used_deployments,
+                        cleanup_failed_tasks: true
                     )
                     # Sanity check that the plan was properly cleaned up
                     SystemNetworkDeployer.verify_all_tasks_deployed(
@@ -406,8 +410,10 @@ module Syskit
                 toplevel_tasks_to_requirements =
                     system_network_generator.toplevel_tasks_to_requirements
 
-                resolution_errors = error_handler.process_failures(
+                resolution_errors, @used_deployments = process_failures(
+                    error_handler,
                     required_instances,
+                    @used_deployments,
                     cleanup_failed_tasks: cleanup_resolution_errors
                 )
                 if cleanup_resolution_errors
@@ -421,6 +427,21 @@ module Syskit
                     )
                 end
                 [required_instances, resolution_errors, toplevel_tasks_to_requirements]
+            end
+
+            def process_failures(
+                error_handler, required_instances, used_deployments, cleanup_failed_tasks:
+            )
+                resolution_errors = error_handler.process_failures(required_instances)
+                return [resolution_errors, used_deployments] unless cleanup_failed_tasks
+
+                removed_tasks = error_handler.cleanup_resolution_errors(
+                    resolution_errors, required_instances, work_plan
+                )
+                removed_tasks.each do |task|
+                    used_deployments.delete(task)
+                end
+                [resolution_errors, used_deployments]
             end
 
             # Computes the system network, that is the network that fullfills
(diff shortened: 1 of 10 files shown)


lib/syskit/network_generation/engine.rb
```diff
@@ -406,8 +410,10 @@ def compute_system_network(
                 toplevel_tasks_to_requirements =
                     system_network_generator.toplevel_tasks_to_requirements
 
-                resolution_errors = error_handler.process_failures(
+                resolution_errors = process_failures(
```
wvmcastro, line 407: I would prefer
```ruby
resolution_errors, @used_deployments = process_failures( ...)

```
because it is very explicit that I am updating the **@used_depoloyments** member

test/network_generation/test_engine.rb
wvmcastro, line 1: I think a few months from now knowing what these tests are _actually_ testing will be a puzzle. I think it is worth documenting what the are testing and how. Moreover, if it wouldn't be too painful to write, I would like to see a test that checks **@used_deployments** before and after error resolution and asserts that the desired deployments were removed

wvmcastro (commented):
Giving our time constraint, I would approve this with just the test documentation

test/network_generation/test_engine.rb
jhonasiv: But they do test a typical behavior from #resolve. The issue before is that we assumed #resolve didnt need any tests due to being tested in higher level elsewhere, but this behavior actually had to be tested since it is specific to this set of parameters. I dont think they are specific to the errors we found, this code path is something that should be directly tested anyways.

Your suggestion is one that is very specific to this bug though, I dont agree that the unit tests should be this granular on this part of syskit.

doudou: I actually prefer the system test, not depending on the internal implementation (which we are fixing actually)

doudou (approved):

lib/syskit/task_context.rb
```diff
@@ -59,6 +59,8 @@ class TaskContext < Component
         # The last state before we went to orogen_state
         attr_reader :last_orogen_state
 
+        attr_accessor :deployed_task
```
doudou, line 62: I would feel better if you would propagate deployed_task during merges (and possibly include it in the merge tests ?)

PR #555, opened 2026-04-10
test(fix): use string as hash keys for conf

wvmcastro:


lib/syskit/test/stubs.rb
```diff
@@ -27,6 +27,11 @@ def self.stub_model_id
 
             # Create empty configuration sections for the given task model
             def stub_conf(task_m, *conf, data: {})
+                # If conf is the configuration hash itself, it will be applied down the
+                # instantiation process by using Orocos::TaskConfigurations#apply,
+                # which accepts the conf Hash directly
+                return unless conf.all? { |x| x.kind_of? String }
```
doudou: `return` ? I do not understand at all.

wvmcastro: If `*conf` is not an array of section names, then it is a Hash (this is the use case I found, thus is not suitable to do `concrete_task_m.configuration_managet.add` beacuse `Orocos::TaskConfigurations#add` expects section name. But this *conf is stored in the task's arguments and will be used later on the [Syskit::TaskConfigurations#apply](https://github.com/rock-core/tools-syskit/blob/d1eae8d9d45c4fc04586929f2ac8a188cb8795c2/lib/syskit/task_configuration_manager.rb#L32) which uses the underlining `Orocos::TaskConfigurations#aply` that accepts Hash.

Maybe I am missing more correct way of working around that, we should discuss

doudou: I think we should.

It was there, it was called, it was not failing

In any case, the absolute rule is: don't ignore a call because you don't like the argument.

lib/syskit/test/stubs.rb
```diff
@@ -27,6 +27,11 @@ def self.stub_model_id
 
             # Create empty configuration sections for the given task model
             def stub_conf(task_m, *conf, data: {})
+                # If conf is the configuration hash itself, it will be applied down the
+                # instantiation process by using Orocos::TaskConfigurations#apply,
+                # which accepts the conf Hash directly
+                return unless conf.all? { |x| x.respond_to?(:to_str) }
+
```
doudou: Just to register our conversion: remove, and fix the tests that are using stub_conf wrong

doudou (approved):

PR #558, opened 2026-05-07
chore: explicitly pass file descriptors to create interfaces in subcommands

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/378

The current method of getting a random port from the kernel and assuming it won't be given to some other process in the time needed to start subcommands is not working in CI with parallel execution. We get a significant number of heisenbugs out of it.

This requires a fork of aruba until the functionality of passing FDs do subcommands gets integrated there.

### DIFF
--- a/test/cli/test_gen_main.rb
+++ b/test/cli/test_gen_main.rb
@@ -8,12 +8,15 @@ module Syskit
         describe "syskit gen" do
             include Roby::Test::ArubaMinitest
 
+            # Override path to CLI command to use ArubaMinitest helpers
+            def roby_bin
+                File.join(Syskit::BIN_DIR, "syskit")
+            end
+
             def assert_app_valid(*args)
-                port = roby_allocate_port
-                syskit_run = run_command(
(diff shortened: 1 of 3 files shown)


jhonasiv (approved):

test/cli/test_gen_main.rb
```diff
@@ -8,12 +8,15 @@ module CLI
...
+                File.join(Syskit::BIN_DIR, "syskit")
+            end
+
             def assert_app_valid(*args)
-                port = roby_allocate_port
-                syskit_run = run_command(
-                    ["syskit", "run", "--port=#{port}", *args].join(" ")
-                )
-                syskit_quit = run_command "syskit quit --retry --host=localhost:#{port}"
+                fileno = roby_allocate_interface_server
```
wvmcastro: I don't like this useless assignment, it makes me think that `roby_allocate_interface_server` did nothing useful. I think just calling it is more informative that it does something under the hood that the reader should go check

wvmcastro (approved):

PR #559, opened 2026-05-08
Fix configuration precedence on replacement

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/380

### DIFF
--- a/test/network_generation/test_engine.rb
+++ b/test/network_generation/test_engine.rb
@@ -348,24 +348,40 @@ module Syskit
 
                         bus_driver.orocos_task.local_ruby_task
                                   .create_output_port("dev", "/int")
-                        flexmock(bus_driver.orocos_task, "bus")
-                            .should_receive(:start).once.pass_thru
-                            .globally.ordered(:bus_startup)
-                        mock_raw_port(bus_driver, "dev")
-                            .should_receive(:connect_to).once.pass_thru
-                            .globally.ordered(:bus_startup)
-                        flexmock(dev_driver.orocos_task, "dev")
-                            .should_receive(:configure).once.pass_thru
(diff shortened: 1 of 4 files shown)


test/test_component.rb
```diff
@@ -452,6 +452,37 @@
...
+            device_driver_m.provides device_m, as: "driver"
+
+            bus = robot.com_bus combus_m, as: "bus"
+            dev = robot.device device_m, as: "dev"
+            dev.attach_to(bus, client_to_bus: false)
+            syskit_stub_configured_deployment(device_driver_m)
+            syskit_stub_configured_deployment(combus_driver_m)
+            dev_driver = syskit_deploy(dev)
+            bus_driver = plan.find_tasks(combus_driver_m).with_parent(dev_driver).first
+            refute dev_driver.meets_configurationg_precedence_constraints?
```
wvmcastro, line 482: Out of curiosity, what happens if `syskit_configure(dev_driver)` is called? It will be a no-op or some kind of exception will be thrown?

test/network_generation/test_engine.rb
```diff
@@ -348,24 +348,40 @@ def deploy_dev_and_bus
...
+                            end
+                        end
+
+                        dev_name = dev_driver.orocos_name
+                        bus_name = bus_driver.orocos_name
+                        expected = [
+                            [:calling, :configure, bus_name],
+                            [:success, :configure, bus_name],
+                            [:syskit_connect, :success,
+                             bus_name, "dev", dev_name, "bus_in", {}],
```
wvmcastro, line 375: I did not follow this call

EDIT: this is the port connection between the bus and the device/driver

wvmcastro (approved):

PR #561, opened 2026-05-14
chore: some ConnectionManagement cleanup

doudou:


jhonasiv (approved):

lib/syskit/runtime/connection_management.rb
```diff
@@ -532,8 +532,13 @@ def mark_connected_pending_tasks_as_executable(pending_tasks)
             end
 
             # Partition a set of connections between the ones that can be
-            # performed right now, and those that must wait for the involved
-            # tasks' state to change
+            # performed right now and those that must be done at the very end of the
+            # modification process
+            #
+            # We perform early changes that involve connections where one of the two tasks
+            # is not running, at the connection is in effect already inactive. When both
```
wvmcastro, line 539: I did not follow "[...] at the connection is in effect already inactive", would it be "[...] as the ..."

lib/syskit/runtime/connection_management.rb
```diff
@@ -542,25 +547,27 @@ def mark_connected_pending_tasks_as_executable(pending_tasks)
...
             def partition_early_late(connections, kind, to_syskit_task)
                 early, late = connections.partition do |(source_task, sink_task), port_pairs|
-                    source_is_running = (syskit_task = to_syskit_task[source_task]) && syskit_task.running?
-                    sink_is_running   = (syskit_task = to_syskit_task[sink_task])   && syskit_task.running?
+                    source_is_running =
+                        (syskit_task = to_syskit_task[source_task]) &&
+                        syskit_task.running?
+                    sink_is_running   =
+                        (syskit_task = to_syskit_task[sink_task]) &&
+                        syskit_task.running?
```
wvmcastro, line 565: safe navigation 
```ruby
source_is_running = to_syskit_task[source_task]]&.running?
sink_is_running = to_syskit_task[sink_task]&.running?
```

wvmcastro (approved):
I have just one suggestion about using safe navigation. And pointed out a typo (probably) The whole think is a lot better.

side note: we could adopt the `refactor:` mark to differentiate this. `chore` is a too big umbrella.

lib/syskit/runtime/connection_management.rb
doudou: Agreed, but since the next PR is rewriting this, WONTFIX.

PR #563, opened 2026-06-02
fix: telemetry UI's state handling

doudou:


lib/syskit/telemetry/async/main_thread_restrictions.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Syskit
+    module Telemetry
+        module Async
+            # Module providing common functionality to validate that some methods are
+            # called within the main thread
```
wvmcastro: I would add a characterization for the "main thread". Something like
```
[...] The main thread is the thread upon the instance was created from
```

lib/syskit/telemetry/async/task_context.rb
```diff
@@ -64,54 +64,43 @@ def states_index_to_symbols
...
-
-                # @api private
+                # This is meant to be called in a separate thread
                 #
-                # Discover a remote task's properties
-                def self.discover_properties(async_task, task)
-                    raw_properties = task.property_names.map { task.property(_1) }
-                    async_task.discover_properties(raw_properties)
+                # @return [Discovered]
+                def self.async_discovery(name, ior, orogen_model)
```
wvmcastro: I can't see how the discovery is being done, it looks like it is just an object initialization

lib/syskit/telemetry/async/main_thread_restrictions.rb
doudou: Updated documentation

lib/syskit/telemetry/async/task_context.rb
doudou: I felt this too, but couldn't put my finger on it. Bad naming. I modified the name of this method (and of its sister from_async_discovery method) to `discover_interface` / `from_discovered_interface`.

wvmcastro (approved):

PR #565, opened 2026-06-12
feat: implement syskit_on_log_rotation hook to trigger rotation on non-orogen loggers

doudou:
Depends on:
- [ ] https://github.com/rock-core/tools-roby/pull/384

### DIFF
--- a/lib/syskit/roby_app/plugin.rb
+++ b/lib/syskit/roby_app/plugin.rb
@@ -205,6 +205,8 @@ module Syskit
                           "cannot enable log transfer without log rotation"
                 end
 
+                Roby.warn "built-in log transfer is deprecated. Use the transfer " \
+                          "functionality of `syskit log_runtime_archive` instead"
                 conf = Syskit.conf.log_transfer
                 conf.target_dir ||= log_dir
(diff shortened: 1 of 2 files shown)


lib/syskit/roby_app/plugin.rb
```diff
@@ -331,6 +331,24 @@ def syskit_log_rotation_poll_handler
...
+                rescue StandardError => e
+                    ::Robot.warn "disabling log rotation handler #{h} because it raised"
+                    Roby.log_exception_with_backtrace(e, ::Robot, :warn)
+                    true
+                end
+            end
+
+            # Register a block that will be called right after logs have been rotated
+            def syskit_on_log_rotation(&block)
+                @syskit_log_rotation_handlers ||= []
```
wvmcastro, line 358: I know it is in the PR title, but I think you should document that these are not orogen logers in the code as well

wvmcastro (approved):

lib/syskit/roby_app/plugin.rb
doudou: It's a bit more fine-grained than that, actually. I significantly expanded the documentation.

PR #566, opened 2026-07-23
add extra CLI commands to syskit

Rezenders:
add `bundle-dir` command to syskit. This command returns the absolute path of a bundle.
Usage example: `syskit bundle-dir my_bundle`

add `syskit config` to export a robot's configuration file.
Usage example: `syskit config my_robot --from-bundle my_bundle --key parent_key.child_key`

Related to: https://github.com/rock-gazebo/simulation-rock_gazebo/pull/70

### DIFF
--- a/lib/syskit/cli/main.rb
+++ b/lib/syskit/cli/main.rb
@@ -80,88 +80,47 @@ module Syskit
                 app.setup_robot_names_from_config_dir
                 app.robot(robot_name)
 
+                merger = proc do |_, old_val, new_val|
+                    if old_val.is_a?(Hash) && new_val.is_a?(Hash)
+                        old_val.merge(new_val, &merger)
+                    elsif old_val.is_a?(Array) && new_val.is_a?(Array)
+                        old_val + new_val
+                    else
+                        new_val
+                    end
+                end
+
                 config_path = ["config", "robots", "ROBOT.yml"]
                 merged_hash = app.find_files(
                     *config_path, all: true, order: :specific_last
-                ).each_with_object({}) do |path, sysdef_hash|
+                ).each_with_object({}) do |path, config_hash|
                     values = YAML.safe_load(File.read(path)) || {}
-
-                    # Define a recursive merging rule
-                    merger = proc do |key, old_val, new_val|
-                        if old_val.is_a?(Hash) && new_val.is_a?(Hash)
-                            # If both are hashes, merge them recursively
-                            old_val.merge(new_val, &merger)
-                        elsif old_val.is_a?(Array) && new_val.is_a?(Array)
-                            # If both are arrays, combine them
-                            old_val + new_val
-                        else
-                            # For simple values the newer file wins
-                            new_val
-                        end
-                    end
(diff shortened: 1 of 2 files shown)


Rezenders:
@jhonasiv @wvmcastro

lib/syskit/cli/main.rb
```diff
@@ -41,6 +41,91 @@ def orogen_test(*args)
...
+
+                bundle = Rock::Bundles.each_bundle.find { |b| b.name == name }
+                if bundle
+                    puts bundle.path
+                else
+                    $stderr.puts "No bundle named '#{name}' found."
+                    $stderr.puts "Available bundles are: #{Rock::Bundles.each_bundle.map(&:name).sort.join(', ')}"
+                    exit 1
+                end
+            end
```
jhonasiv, line 56: This possibly exists in autoproj, or at least it is possible to get this information from an autoproj command (ie autoproj show bundles/bla)

lib/syskit/cli/main.rb
```diff
@@ -41,6 +41,91 @@ def orogen_test(*args)
...
+
+            desc "config ROBOT", "print the configuration of a given robot"
+            option :key, type: :string, repeatable: true, default: []
+            option :from_bundle, type: :string, default: nil
+            def config(robot_name)
```
jhonasiv, line 61: Typically, syskit is ran from the bundle itself. Besides that, I think a more useful command would to extract the variables inside the `Conf` variable from a syskit run, but that is way more complex (and might encounter friction if `Conf` is used as something dynamic). 

I dont think that what you aimed for here should be packed with syskit, it can mostly be resolved by using a sequence of bash commands, and it is outside of syskit's scope

jhonasiv (changes requested):

lib/syskit/cli/main.rb
Rezenders: Yes, you can find packages with `autoproj locate PACKAGE_NAME` or with `autoproj show PACKAGE_NAME` as you said. But IMO it can be helpful to have this extra command, it is simple, easy to remember, and doesn't require autoproj. On the other hand, it doesn't bring any disadvantages

lib/syskit/cli/main.rb
Rezenders: > I think a more useful command would to extract the variables inside the Conf variable from a syskit run, but that is way more complex (and might encounter friction if Conf is used as something dynamic).

I could also that, but I believe it should be a distinct command from this one. The current proposed `syskit config` command is designed to work without requiring the system to be running, which is nice to have. Otherwise, one would need to start the system to inspect the config files. 
A part from the templated sdf generation use case we have in mind, this command could be useful for a quick inspection on the configuration of the system without needing to open all the files manually and read them, which is convenient.
Again, it doesn't bring any disadvantages adding support for this command. 

> syskit is ran from the bundle itself.

Yes, but for the templated SDF generation use case we dont run `rock-gazebo` necessarily from the robot bundle dir, we could run it from an arbitrary directory.

Rezenders:
@doudou

